PR #221, opened 2026-05-03
feat: status-bar indicator for large file transfers (#127)

### LINKED ISSUE #127: UX: progress indicator for large file uploads / downloads (>1 MB)
## Symptom

When the user (or a plugin) reads/writes a file >1 MB through the patched adapter, the operation can take seconds-to-tens-of-seconds over a typical SSH RTT × payload size product. Today the UI is silent during that wait — no spinner, no byte counter, no Notice. Users assume the editor froze and Cmd-Q out of frustration.

Reviewer flagged: "10MB以上のアップロード/ダウンロード中に進捗バーがない".

## Where it's missing

- `SftpDataAdapter.writeBuffer` — single `client.writeBinary` call, no chunking visibility.
- `SftpDataAdapter.readBuffer` — same on the read side.
- `RpcRemoteFsClient.writeBinary` / `readBinary` — single RPC call with the whole base64 payload.
- `ResourceBridge` — image / PDF / video served on every range request, no progress on the initial fetch.

## Proposed shape

1. **Add a `progressCallback?: (bytesTransferred, totalBytes) => void` option** to:
   - `SftpDataAdapter.write` / `writeBinary` / `read` / `readBinary` — passed through from Obsidian's adapter call.
   - `RpcRemoteFsClient.writeBinary` / `readBinary` — emit between SSH frame writes (need a chunk-emitting variant; current API single-shots a base64 string).

2. **Surface in the StatusBar** — when ≥1 file with `totalBytes > 1 MB` is in flight, show `↓ 4.2 / 12.0 MB · note.pdf` with a progress bar. Multiple concurrent ops collapse to `↓ 18 MB · 3 files`.

3. **Threshold** — only show the indicator for ops > 1 MB (default; settings-overridable). Sub-MB ops are <500 ms and a transient indicator would be more visual noise than help.

## Open questions for the design pass

- The current RPC frame is "send the full base64 in one JSON value, get the full result". Adding chunked progress means either (a) a separate `fs.writeBinaryStream` proto method that takes a sequence of chunks, or (b) approximating progress from upload byte counts at the SSH framing layer (less accurate but no proto change).
- Approach (a) couples nicely with `fs.readBinaryRange` (#new-issue-from-reviewer's-list, pending architect triage) — both want a streaming wire shape.
- For SFTP transport, the underlying `ssh2` library does emit byte-progress on its WriteStream — wire that up directly.

## Done when

- [ ] Writing or reading a 10 MB file shows a StatusBar progress indicator that updates ≥4× during the transfer.
- [ ] The indicator goes away within 500 ms of the last in-flight op completing.
- [ ] Plugin-driven ops (Templater bulk file create, Excalidraw 5 MB image embed) get the indicator the same way.
- [ ] Sub-1 MB ops produce no UI.

## Cross-refs

- Plan: §F (UX phase) — fits Phase F-G UX work.
- Issue #126 — Phase F-H roadmap; this expands F's UX deliverables.
- Pending architect triage on `fs.readBinaryRange` (B-group from reviewer review session) — if streaming wire shape is chosen, this issue's RPC-side option (a) lands as part of that work instead of separately.

sotashimozono:
## Summary

Show a status-bar indicator when uploads or downloads larger than 1 MB are in flight, so users see something is happening during slow transfers instead of assuming the editor froze.

Format: `↑ 5.2 MB · note.pdf · 3s` (single transfer) / `↕ 18.0 MB · 3 files · 5s` (multiple).

### What's in
- `TransferTracker` — in-memory event emitter with `begin`/`end`/`subscribe`/`clear`/`snapshot` (>1 MB threshold)
- `LargeTransferBar` — status bar item with 250 ms tick for elapsed seconds, hidden when nothing in flight
- `SftpDataAdapter.writeBuffer` wraps `client.writeBinary` in begin/end
- `SftpDataAdapter.readBuffer` wraps the cache-mismatch re-read path (uses the stat we already pay for to know size upfront)
- `AdapterManager.restore()` calls `transferTracker.clear()` to drop stale visualizations on disconnect
- 12 unit tests for `TransferTracker`
- CSS for tabular numerals + hidden state

### What's deferred
True chunked progress (≥4 updates per transfer, per the issue's "Done when") needs wire-protocol changes (`fs.writeBinaryStream` proto method) and is tracked separately. This ship eliminates the "editor froze, Cmd-Q" pain without protocol churn — the elapsed-seconds counter advancing is enough to distinguish "in progress" from "stuck".

Cold-start large reads (no cache entry yet) don't get the indicator since we don't know the size upfront and the issue notes that big binaries usually flow through `getResourcePath` → `ResourceBridge` (which has its own range-fetch path).

### Side fix
Removed accidentally committed `.claude/PRPs/reviews/pr-*.md` files (per-developer Claude Code state) and added `.claude/` to `.gitignore`.

## Test plan
- [x] `npx tsc --noEmit` — zero errors
- [x] `npx vitest run` — 45 files, 663 tests pass (12 new TransferTracker tests)
- [x] `npx eslint` on touched files — clean (the 16 pre-existing AdapterManager errors are not introduced here)
- [ ] Manual: write a 5 MB image to vault, verify bar appears with elapsed counter

Closes #127


sotashimozono (commented):
## Self-review (Claude Opus 4.7)

**Decision:** APPROVE — code is small, mirrors existing patterns (PendingEditsBar, AncestorTracker), well-tested at the tracker level, CI is green.

### Findings: 0 critical, 0 high, 2 medium, 3 low

**MEDIUM**
1. **Missing test coverage for LargeTransferBar.** The DOM-bound render/tick logic has no unit test. Mocking `activeWindow` and `Plugin.addStatusBarItem()` is feasible — would prevent silent regressions on a future refactor of the show/hide cycle.
2. **No integration test for SftpDataAdapter wiring.** The new `transferTracker?.begin/end` calls in `writeBuffer`/`readBuffer` aren't asserted by `tests/SftpDataAdapter.test.ts`. A test passing a fake tracker and asserting begin/end fire on a >1 MB write would document the contract.

**LOW**
1. `LargeTransferBar.ts:78` uses `Math.min(...transfers.map(...))` — would blow the call stack at ~10⁵ entries. Use `reduce` for safety: `transfers.reduce((m, t) => Math.min(m, t.startedAtMs), Infinity)`. Not realistic for our use case but trivial to harden.
2. `LargeTransferBar.ts:30` magic number `250` — a named `TICK_INTERVAL_MS` constant would be self-documenting.
3. `SftpDataAdapter.ts:721` — `s.size ?? 0` silently treats 0-byte stat as "no transfer" because `begin()` rejects sub-threshold sizes. Correct behavior, worth a one-line comment.

### Validation
- ✅ `tsc --noEmit` clean
- ✅ `eslint` clean (after the AdapterManager `as any` fix in this PR)
- ✅ `vitest run` — 45 files / 663 tests pass (12 new for TransferTracker)
- ✅ All required CI: lint, typecheck, server build, SSH integration, security scans

### Scope note
Issue #127 §"Done when" says "≥4 updates per transfer". This PR deliberately defers true chunked progress (needs wire-protocol changes — `fs.writeBinaryStream` proto method) and ships only start/end + elapsed seconds. That's enough to solve the actual UX pain ("user assumes editor froze") without protocol churn.

Issue #130, opened 2026-04-29
UX: remote vault path browser modal (no more typing paths by hand)


sotashimozono:
## Symptom

To configure a profile today, the user has to type the **remote vault path** by hand into a text input. They need to know:

- whether their vault sits at `~/notes/main` or `/srv/obsidian/main` or somewhere else
- whether `~` works (it does — `resolveRemotePath` expands it)
- whether to write a trailing slash (it doesn't matter, but no UI tells them)

Reviewer flagged: "SSHでリモートFSを参照してVaultフォルダを選ぶUIが欲しい".

## Proposed shape

A "Browse remote…" button next to the **Remote vault path** field in `ProfileForm.ts`, enabled only when:

- `host` + `port` + `username` + auth fields look complete (so we can actually open a session), AND
- A test SSH connection to the profile succeeds (or one is already open via a sibling connect button).

Click → modal opens with:

1. **Breadcrumb** at the top (`~ / notes / main`).
2. **Tree view** of the current directory's entries — folders only by default, files behind a "show files" toggle.
3. **"Use this folder" button** at the bottom — fills the parent text input + closes modal.
4. **Cancel button** — closes without changing anything.
5. **"New folder" button** — `mkdir -p` shortcut for the case where the user hasn't created the vault dir yet.

## Implementation notes

- Folder listing reuses the daemon's `fs.list` (RPC) or `SftpClient.list` (SFTP) — the existing transport layer already gives us a `RemoteEntry[]` per dir.
- For non-RPC profiles (SFTP-only), this works the same — `SftpClient.list` is the path.
- Initial cwd: the remote `$HOME` (already resolved by `getRemoteHome()` in `DeployerSshClient`). Lets the user navigate down from a known starting point.
- Path normalization: store the user's selection as `~/relative/path` if it's home-rooted (preserves the user's mental model), otherwise as the absolute path.
- **Don't** open a long-lived SSH session for the modal — open just-in-time, close on modal dismiss. Reusing the active connect's session is fine if one is already up.

## What this is NOT

- Not a vault initializer — picking a folder doesn't create or import vault content. Just sets the profile's `remotePath`.
- Not a security boundary — the daemon's vault-root sandbox stays in effect (`PathOutsideVault` errors still possible if the user picks a path that doesn't agree with the daemon's start arg).

## Done when

- [ ] Profile form has a "Browse remote…" button.
[...]

Issue #126, opened 2026-04-29
Phase F → G → H roadmap: UX rework → BRAT beta → v1.0 + community store registry


sotashimozono:
## Goal

The three remaining roadmap phases between Phase E (compat, #124) and v1.0. Aggregated here as a single tracking issue because they're tightly sequenced — F lands the UX needed for G's beta, G's beta surfaces feedback that informs H's registry-submission polish.

Plan: `C:/Users/souta/.claude/plans/obsidian-remote-ssh-composed-sunset.md` §A — table at "Phase ロードマップ".

---

## Phase F — UX polish (P0 path to community-store-quality UX)

Exit criteria: Onboarding produces a working profile in <5 clicks; status-bar shows queue depth; `tests/ui/` + `tests/settings/` coverage to ≥60 %.

- [ ] **F15 — Settings UI rework** — `plugin/src/settings/SettingsTab.ts` + `ProfileForm.ts`. Profile list with per-profile health badge (connected / disconnected / error). Secret form using OS keychain via `SecretStore.ts`. Currently P0 / partial.
- [ ] **F16 — Status bar + command palette** — `plugin/src/ui/StatusBar.ts`. Surface: connect/disconnect/queue-depth/cancel-reconnect. Command palette aliases for the same. Currently P0 / partial.
- [ ] **F17 — Onboarding wizard** — new `plugin/src/ui/OnboardingModal.ts`. First-launch profile builder: detect `~/.ssh/config`, key gen if no key, sshd connectivity hint. Goal is "<5 clicks to working profile". Currently P1 / new.
- [x] **UI test coverage** — shipped in #229 + follow-up #230. `tests/__mocks__/obsidian.ts` runtime mock; `src/ui/**` + `src/settings/**` no longer excluded from coverage; OnboardingModal.ts at 66.66% lines / 47.36% branches, SettingsTab.ts at 53.33% / 60%.

Estimated PR count: 6-10.

---

## Phase G — BRAT beta + telemetry opt-in

Exit criteria: 2 weeks of BRAT distribution with no P0 bugs; ≥10 external installs.

- [x] **`manifest-beta.json`** — shipped in #231. `manifest-beta.json` lives at repo root, `bump-version.mjs` keeps it in sync, `release.yml` attaches it as a release asset, `version-check.yml` enforces parity. README split into Option A (BRAT) / Option B (manual).
[...]

PR #222, opened 2026-05-03
feat: first-launch onboarding wizard (#126 F17)

sotashimozono:
## Summary

Phase F17 — onboarding wizard. New users get a single-modal flow from "no profiles" to "ready to connect" in <5 clicks when an `~/.ssh/config` host already exists.

### Flow

1. **Host details** — auto-import dropdown when `~/.ssh/config` has entries; manual fields for everything
2. **Auth** — detect existing `~/.ssh/id_ed25519` / `id_rsa`; one-button ed25519 keygen (mode 0600 private, 0644 public — OpenSSH-strict) + copy-public-key button
3. **Remote vault path** — text input + Browse button reusing `RemotePathBrowserModal` (#130)
4. **Save & test connection** — runs an `SftpClient` smoke connect+list and persists the profile only on success

Trigger: opens automatically in `onLayoutReady` when `profiles.length === 0 && !onboardingCompleted` and the vault is not a shadow-vault auto-connect target. Modal dismiss flips `onboardingCompleted` so it doesn't reopen. Re-launchable from the command palette as `Set up first remote vault`.

### New files

| File | Purpose |
|---|---|
| `plugin/src/ssh/SshKeyGen.ts` | `generateEd25519KeyPair(path, comment)` + `pemPublicKeyToOpenSshEd25519()` helper that re-packs SPKI into OpenSSH wire format. No `sshpk` dependency. |
| `plugin/src/ui/OnboardingModal.ts` | The wizard. |
| `plugin/tests/SshKeyGen.test.ts` | 6 tests: wire-format round-trip, RSA-rejection, file write, parent-dir mkdir, key uniqueness, mode 0600/0644 (skipped on Windows). |

### Wire-up
- `PluginSettings.onboardingCompleted?: boolean` (optional, defaults falsy → modal opens on first launch)
- `main.ts` `onLayoutReady` — trigger condition + skip on shadow vaults
- `main.ts` command `show-onboarding` — re-launchable

### Click count to working profile

When user has `~/.ssh/config` and existing key:
1. Pick host from SSH config dropdown — fills host/port/user/key (1)
2. Click Browse… → pick folder → Use this folder (2)
3. Save & test connection (1)

= **4 clicks**. Meets the Phase F exit criterion of <5 clicks.

## Test plan
- [x] `npx tsc --noEmit` — zero errors
- [x] `npx eslint` on touched files — clean
- [x] `npx vitest run` — 46 files / 668 pass / 1 skipped (Windows perm bits)
- [ ] Manual: install on a fresh vault, verify modal opens, generate key, smoke-test against a real sshd

Closes the F17 checkbox in #126


sotashimozono (commented):
Code review summary — **APPROVE with comments**. No critical/high blockers; SPKI→OpenSSH wire-format conversion is correct and well-tested, wizard wires cleanly into the existing deps/modals, and the trigger condition is right (after the shadow-vault branch, only when `profiles.length === 0 && !onboardingCompleted`).

Validation: `tsc --noEmit` clean, `vitest run tests/SshKeyGen.test.ts` 5 pass / 1 Windows-skip.

### Medium
- **M1** `OnboardingModal.onClose` — `void this.markCompleted()` swallows errors. If `saveSettings` ever rejects the modal will reopen forever with no log. Add a `.catch(e => logger.error(...))`.
- **M2** Save-and-test happy path writes `data.json` twice in a row (`onSave` then `markCompleted`). Coalesce by setting both fields before a single `saveSettings`.

### Low
- **L1/L2** `applySshConfig` and the Browse callback both call `renderBody()`, so the SSH-config dropdown selection visually resets and the public-key hint box gets nuked. Update individual `Setting`s imperatively, or persist `selectedAlias` on `this`.
- **L3** Hardcoded fallback filename `id_ed25519_obsidian_remote` in `generateKey` — lift to `constants.ts`.
- **L4/L5** TOCTOU between `existsSync` and `writeFile`, plus `writeFile({ mode })` only sets perms on creation. Switching `generateEd25519KeyPair` to `fs.open(path, 'wx', 0o600)` makes both safe by syscall and lets you drop the modal-side existsSync guard.
- **L6** `import * as crypto from 'crypto'` in `OnboardingModal` — only used for `randomUUID()`; `globalThis.crypto.randomUUID()` is portable and avoids a node-only import. Style nit.
- **L7** The "rejects non-ed25519" test generates a real RSA-2048 key (~100ms) just to fail the length check. A small hand-rolled non-44-byte SPKI buffer would be faster and equally clear.

Full report: `.claude/PRPs/reviews/pr-222-review.md`

sotashimozono:
Pushed cda3dac addressing the review:

**Medium (must-have)**
- ✅ **M1** `OnboardingModal.onClose` now `.catch(e => logger.error(...))` so a `markCompleted` rejection surfaces in the log instead of silently re-opening the wizard on every launch.
- ✅ **M2** Coalesced double `saveSettings`. The constructor now takes a single `onFinish({profile?, dismissOnboarding})` callback and `main.ts` does one disk write per save-and-test path.

**Low**
- ✅ **L3** Hardcoded fallback filename → \`constants.ONBOARDING_FALLBACK_KEY_FILENAME\`.
- ✅ **L4/L5** \`generateEd25519KeyPair\` now uses \`fs.open(path, 'wx', 0o600)\` for atomic exclusive-create. Closes the TOCTOU window AND ensures mode 0o600 is set at create time (\`writeFile({mode})\` was a no-op when the path already existed). Dropped the modal-side \`existsSync\` guard. New unit test asserts pre-existing content survives an EEXIST attempt.
- ✅ **L7** Test no longer generates a real RSA-2048 key (~100 ms); uses a hand-rolled 64-byte buffer that triggers the same length guard.
- ⏭️ **L6** Reconsidered: \`ProfileForm.ts:31\` uses node \`crypto.randomUUID\` (project convention) and the \`obsidianmd/prefer-active-doc\` lint rule fires on \`globalThis\`. Kept node crypto for consistency.
- ⏭️ **L1/L2** Skipped — re-rendering on \`applySshConfig\` and the Browse callback is rare enough that the code-cost of incremental Setting updates outweighs the UX benefit. Worth revisiting if user reports drift.

Validation: \`tsc --noEmit\` clean, \`eslint\` clean, \`vitest run\` 46 files / 669 tests pass / 1 Windows-skipped.

PR #223, opened 2026-05-03
feat: opt-in anonymous telemetry (#126 F22)

sotashimozono:
## Summary

Phase F22 — opt-in anonymous telemetry. Counts error categories and reconnect-state outcomes locally so users (and BRAT bug-bash testers) can paste a histogram into a bug report without leaking vault content. **Nothing is ever sent over the network.**

### Surface

- **\`Telemetry\` singleton** in \`util/Telemetry.ts\` — \`recordError(category, code)\` / \`recordReconnect(state)\` no-ops while disabled, in-memory \`Map<key, count>\` while enabled.
- **Observation hooks** — single-funnel:
  - \`classifyToNotice\` (the only path that turns thrown errors into Notices) calls \`recordError(classified.category, classified.code)\`.
  - \`onReconnectStateChange\` calls \`recordReconnect(s.kind)\`.
- **Persistence** — JSONL append-log at \`<vault>/.obsidian/plugins/remote-ssh/telemetry.jsonl\`. 60s periodic flush + flush on disable. Counters reset on every flush — file is a stream of deltas, never grows unbounded.
- **Settings UI** — Telemetry section with toggle + View / Flush now / Reset buttons. View opens a copy-to-clipboard modal (the BRAT-tester-friendly format).
- **Default off**. Both observation hooks early-return on disabled, so steady-state cost is ~zero.

### Privacy
JSONL field shape:
\`\`\`json
{"at": 1714712345678, "kind": "error", "category": "timeout", "code": -32000, "count": 3}
{"at": 1714712405678, "kind": "reconnect", "state": "recovered", "count": 1}
\`\`\`
No path / host / user / vault content / stack traces. Only enum keys + integer counts.

### Tests
16 unit tests covering: disabled-default no-op, setEnabled preconditions, idempotent enable, counter aggregation, distinct keys, non-numeric codes, JSONL append shape, append-not-overwrite, no-op flush when nothing buffered, reset without flush.

## Test plan
- [x] \`tsc --noEmit\` clean
- [x] \`eslint\` clean
- [x] \`vitest run\` — 47 files / 685 tests pass / 1 Windows-skipped
- [ ] Manual: enable in settings, trigger a connect failure, verify counter appears in View modal and is appended to telemetry.jsonl on flush

Closes the F22 checkbox in #126


sotashimozono (commented):
Code review — **APPROVE with comments**. No critical/high blockers. Singleton is disabled by default (all hooks are no-ops until opt-in), 16/16 tests pass, tsc clean.

### Medium

**M1 — `counters.clear()` before `appendFile` loses data on write failure**
In `Telemetry.ts flush()`, the map is cleared before the async write. If `appendFile` fails (disk full, permissions), all buffered counts for that interval are silently lost. Fix: move `counters.clear()` inside the try block, after the successful write.

**M2 — Path construction duplicated in two places**
`main.ts` and `SettingsTab.ts` both build the same `path.join(basePath, configDir, 'plugins', manifestId, 'telemetry.jsonl')`. Extract to a helper so it's one source of truth.

### Low

**L1** — `setEnabled(true, newPath)` silently ignores `newPath` if already enabled (early-return guard). Not a live bug since both callers use the same path, but the API looks like it supports re-pointing. Add a JSDoc note.

**L2** — `snapshot()` sets `at: Date.now()` (snapshot time), but `TelemetryRecord.at` JSDoc says "when the record was flushed". Correct for flush records, wrong for snapshot records. Since `at` isn't shown in the UI this is invisible to users, but worth a doc fix.

**L3** — `void telemetry.setEnabled(false)` in `onUnload` is fire-and-forget — inherent to the plugin lifecycle, acceptable by design.

**L4** — `timer: number | null` vs Node's `NodeJS.Timeout` — intentional per the comment, but worth a note that the `number` annotation is Obsidian-specific and depends on the vitest shim.

Full report: `.claude/PRPs/reviews/pr-223-review.md`

sotashimozono:
Pushed e76c4c3 addressing the review:

**Medium**
- ✅ **M1** \`flush()\` now clears counters only after \`appendFile\` resolves. Write failure (disk full / permissions) leaves buffered deltas intact for the next attempt instead of silently dropping data. New unit test forces EISDIR via a directory-as-path and asserts the counts survive.
- ✅ **M2** Extracted \`telemetryLogPath(basePath, configDir, manifestId)\` helper exported from \`util/Telemetry.ts\`. \`main.ts\` (onload) and \`SettingsTab.ts\` (toggle handler) both call it. New unit test asserts the join shape.

**Low**
- ✅ **L1** \`setEnabled\` JSDoc clarifies idempotency: a second \`setEnabled(true, newPath)\` ignores \`newPath\`. To re-point the log, disable then re-enable. Plugin always uses \`telemetryLogPath\` so this hasn't been an issue in practice.
- ✅ **L2** \`snapshot()\` JSDoc clarifies that \`at\` is snapshot time (not original event time) — counters are deltas, per-event timestamps were never tracked.
- ⏭️ **L3** Fire-and-forget on \`onunload\` accepted as documented (plugin lifecycle).
- ⏭️ **L4** \`timer: number | null\` accepted as documented (Obsidian-specific, vitest shim covers tests).

Validation: \`tsc --noEmit\` clean, \`eslint\` clean, \`vitest run\` 47 files / 687 tests pass / 1 Windows-skipped (added 2 new tests).

Issue #124, opened 2026-04-29
Phase E: plugin compatibility — harness foundation + 4 plugin scenarios


sotashimozono:
## Goal

Verify the most-installed Obsidian community plugins (Dataview / Templater / Excalidraw / Tasks) work end-to-end against a shadow vault driven by this plugin. Per the master plan §A.E:

> 10–15 PR | 中リスク (Obsidian internal API drift)
> Exit criteria: DV/Templater/Excalidraw/Tasks の scripted シナリオが nightly で green

## Harness decision (the first sub-task)

Phase E's biggest design choice is **how** to test plugin compat. Four candidates, ranked:

| Approach | Cost | Fidelity | Notes |
|----------|------|----------|-------|
| **A. Playwright + real Obsidian Electron** | Large (~150 MB bundle / nightly only) | Highest (real DOM + obsidian.js) | Deferred from Phase C M11; tracked separately in its own issue. |
| **B. obsidian-mock + scripted scenarios** | Small (vitest unit) | Medium (mock has limits) | Extends M9's FakeVault; duck-types each plugin's hot APIs. |
| **C. Manual verify + `docs/plugin-compatibility.md` updates** | Trivial | Low (no regression detection) | Already partially done; not a substitute for automation. |
| **D. CDP (`--remote-debugging-port=9222`) tail of a running Obsidian** | Medium | High (real binary, low reproducibility) | Memory has a CDP-tail script note; useful for souta's manual loop, not for CI gating. |

**Recommended start: B**, beginning with F11 Dataview as the first scenario. If B's mock turns out to miss critical paths a real plugin hits, *then* invest in A (M11).

## Sub-tasks (PR-sized)

- [ ] **E-α: compat harness** — `plugin/tests/compat/` directory; FakeVault extended with Obsidian's hot read APIs (`vault.getMarkdownFiles`, `metadataCache.getFileCache`, `vault.cachedRead`, `vault.read`, `app.workspace.activeLeaf`). Picks B from the matrix above. Documents the duck-type surface so subsequent plugin cases plug in.

- [ ] **F11 — Dataview compat smoke** — extend the harness with a 5-file fixture vault, run the Dataview-equivalent of `dv.pages('"folder"')` (i.e. drive the public `metadataCache.getFileCache()` for each file in `vault.getMarkdownFiles()`), assert all 5 paths + frontmatter come back. The smoke test of "Dataview can list our shadow vault."

[...]

PR #224, opened 2026-05-03
test: Phase E compat harness foundation (#124 E-α)

sotashimozono:
## Summary

Phase E E-α — the foundation subtask of #124. Adds a duck-typed Vault + MetadataCache (Approach B from the issue matrix) so subsequent plugin scenarios can run without a real Obsidian Electron process. Fast, deterministic, vitest-native.

### What's in
- **\`CompatVault.ts\`** — \`CompatVault\`, \`CompatTFile\`, \`CompatTFolder\`, \`CompatMetadataCache\`. Hot duck-type surface identified from the docs/plugin-compatibility.md top-20 survey:
  - \`vault.getMarkdownFiles / getAbstractFileByPath / read / cachedRead / readBinary / create / createBinary / modify / delete\`
  - \`vault.on / offref / trigger\` events (\`create | modify | delete | rename\`)
  - \`metadataCache.getFileCache(file)\` returning \`CachedMetadata\` with \`frontmatter\` (YAML scalars), \`headings[]\` (1-6 + text), and \`listItems[]\` (checkbox tasks open vs done)
- **\`fixtures.ts\`** — \`loadFixtures(vault, dir)\` walks for \`.md\` files; \`fixturesDir()\` resolves to the bundled fixtures dir
- **4 fixtures**: \`note-with-frontmatter.md\`, \`note-with-headings.md\`, \`note-with-tasks.md\`, \`plain-note.md\` (negative-control)
- **\`harness.test.ts\`** — 11 smoke tests covering the full surface

### Documented in \`docs/plugin-compatibility.md\`
New "Compat harness (#124 Phase E)" section with:
- File map
- Hot duck-type surface table
- "Adding a new scenario" recipe so F11-F14 PRs can plug in directly

### What's NOT in
F11 (Dataview), F12 (Templater), F13 (Excalidraw), F14 (Tasks), and the nightly compat workflow land as **separate follow-up PRs** per the issue's sequencing.

### Stats
8 new files, +561 lines. \`vitest run\` 48 files / 698 pass / 1 Windows-skipped (11 new harness tests).

## Test plan
- [x] \`tsc --noEmit\` clean
- [x] \`vitest run tests/compat/\` — 11/11 pass
- [ ] F11 (Dataview) PR will exercise the harness end-to-end

Ref #124 (E-α only — F11-F14 in separate PRs)


sotashimozono (commented):
Code review — **APPROVE with comments**. No critical/high blockers. Clean test-infrastructure PR; tsc clean, 11/11 harness tests pass.

### Medium

**M1 — `void vault.create(rel, body)` in `fixtures.ts`**
Voids an async call. Works because `CompatVault.create` is synchronous under the hood today, but fragile:
- If `create` rejects (duplicate path from a fixture), error is swallowed and the returned count is wrong.
- If `create` is ever refactored to be truly async, the vault state may not be ready when `loadFixtures` returns.

Fix: make `loadFixtures` async and `await vault.create(...)` inside the loop. The `beforeEach` in `harness.test.ts` is already async so no other change needed.

**M2 — `getFileCache` JSDoc says "Compute (and memoise)"**
`CompatMetadataCache.getFileCache` only reads from the cache; the computation is in `rebuildFor`. The comment is misleading — update to "Return the memoised CachedMetadata for `file`, or null if not cached."

### Low

**L1 — `vault!: Vault` definite assertion is never fulfilled**
Both `CompatTFile` and `CompatTFolder` declare `vault!: Vault` but the field is never assigned. The `!` tells TypeScript it's safe to read but it's always `undefined`. Intentional (plugins that access `file.vault` are out of harness scope), but add a comment to that effect so future readers don't treat it as a bug.

**L2 — `as never` casts in `harness.test.ts`**
`file as never` used ~8 times to bridge `TAbstractFile | null` to vault APIs. A one-liner `function asFile(f) { return f as unknown as TFile; }` in the test file would document the intent and reduce noise.

**L3 — `FRONTMATTER_RE` doesn't match `\r\n` line endings**
`\n---\n?` won't match Windows-authored fixtures. Current bundled fixtures are LF so no issue now. A `.replace(/\r\n/g, '\n')` normalizer at the top of `parseFrontmatter` would make it robust.

Full artifact: `.claude/PRPs/reviews/pr-224-review.md` (to be written)

sotashimozono:
Pushed e97696f addressing the review:

**Medium**
- ✅ **M1** \`loadFixtures\` is now async + awaits each \`vault.create\`. Drops the \`void\` cast that swallowed rejections (e.g. duplicate path) and decouples the harness from CompatVault.create's current synchronous internals. \`harness.test.ts\` beforeEach already awaits.
- ✅ **M2** \`CompatMetadataCache.getFileCache\` JSDoc clarifies it's a pure read; computation lives in \`rebuildFor\`.

**Low**
- ✅ **L1** Added a comment on \`CompatTFile.vault!\` / \`CompatTFolder.vault!\` — the \`!\` is "don't surface as a typing error", not "promise this is initialised". Plugins touching \`file.vault\` are out of harness scope.
- ✅ **L2** New \`asFile(TAbstractFile | null)\` helper in \`harness.test.ts\`. Replaced 12 \`as never\` casts with one named cast that throws on null + documents intent. Only the comment reference remains.
- ✅ **L3** \`parseFrontmatter\` normalises CRLF → LF. New unit test asserts a CRLF body round-trips frontmatter + headings the same as LF.

Validation: \`tsc --noEmit\` clean, \`vitest run\` 48 files / 699 pass / 1 Windows-skipped (12 harness tests, +1 CRLF).

Issue #170, opened 2026-04-30
Implement basePath + getBasePath patching in PATCHED_METHODS (#133 follow-up)


sotashimozono:
The #133 basePath compat survey (PR #165, see `docs/plugin-compatibility.md`'s "basePath compat survey" section) established that 6 / 20 top-installed plugins read `app.vault.adapter.basePath` (or the more common `getBasePath()` method). Three are high-risk fs-read (Templater, Kanban, Importer); the survey concluded the high-leverage fix is a single uniform patch returning the shadow-vault local path.

This issue is the implementation. The survey itself shipped no code change.

## What ships

- Extend `PATCHED_METHODS` ([plugin/src/main.ts:49](plugin/src/main.ts#L49)) to include `basePath` and `getBasePath`. Both forms route to the shadow-vault local root (which is already a real local directory mirrored by the file-watcher).
- Unit test on `AdapterPatcher` confirming both the property getter and the method form are intercepted and return the shadow-vault path.
- Update `docs/plugin-compatibility.md` to mark Templater, Kanban, Importer, Copilot as `verified` once smoke tests pass in dev vault.

## What does NOT ship in this issue

- obsidian-Git: shells out to local `git` binary; no fix at the basePath layer. Document as known-incompatible (already noted in the survey).
- `app://local/<path>` URL rewriting (separate issue — plugins that build asset URLs by hand bypass `getResourcePath`).

## Done when

- Templater's `tp.file.path(false)` resolves to a path that, when read, returns the synced shadow-vault content.
- Kanban's clipboard image-paste flow lands the file in the shadow vault and propagates to the remote.
- Importer's Evernote → MD output appears on the remote after the import.

## Estimated scope

Small — ~50 LOC plus tests. Survey already decided the design.

## Cross-refs

- #133 (parent investigation issue)
- PR #165 (survey deliverable)
- `plugin/src/main.ts:49` `PATCHED_METHODS` declaration

PR #226, opened 2026-05-03
test: F12 Templater compat scenario (#124)

sotashimozono:
## Summary

Phase E F12 — Templater compat scenario on top of the harness. Simulates Templater's \`tp.file.create_new(template, filename)\` + subsequent \`vault.modify()\` flow against \`CompatVault\`.

The point is **NOT** to embed Templater itself — Templater needs no work from us, just the public Vault APIs. The point is to verify those round-trip the way Templater expects.

### Hot APIs exercised
- \`vault.read(template)\` — load template body
- \`vault.create(filename, content)\` — drop the rendered file
- \`vault.modify(file, content)\` — user edits the result
- \`vault.on('create' | 'modify')\` — Templater listens for both
- \`metadataCache.getFileCache(...)\` — frontmatter shows up post-create

### Files
- 3 templates under \`fixtures/templater/Templates/\`: \`daily-note.md\` (date placeholder), \`project.md\` (frontmatter + \`tp.file.title\`), \`meeting.md\` (no-date negative-control)
- \`templater.test.ts\` with 6 tests:
  - daily renders \`<% tp.date.now %>\` against a frozen clock
  - project renders frontmatter + \`tp.file.title\` heading; \`metadataCache\` reflects new frontmatter immediately
  - meeting renders without any date placeholder
  - duplicate target path rejection
  - missing template rejection
  - \`vault.modify\` round-trips body + cache refresh

The \`<% expr %>\` evaluator is a deliberately-tiny substitute for Templater's real one — just enough to cover \`tp.date.now(fmt)\` and \`tp.file.title\` (the only placeholders the bundled templates use).

### Doc
\`docs/plugin-compatibility.md\` — Templater row flipped from "🟡 expected (smoke pending after #170)" to **"✅ verified-by-harness (#124 F12)"**.

### Stats
5 files changed, +215 lines. \`vitest run\` 50 files / 713 pass / 1 Windows-skipped.

### What's next
F13 Excalidraw, F14 Tasks, and the nightly compat workflow remain as follow-up PRs in #124.

## Test plan
- [x] \`tsc --noEmit\` clean
- [x] \`vitest run tests/compat/\` — 26/26 (12 harness + 8 dataview + 6 templater)

Ref #124 (F12 only)


sotashimozono (commented):
APPROVE — tsc clean, 26/26 tests pass (harness + dataview + templater), no production code changes.

Frozen-clock pattern is clean; `formatDate` using UTC methods is correct. `evalTemplate` throwing on unknown expressions is the right call for a deliberately-scoped harness.

### Low (advisory, no blockers)

**L1 — `attendees: []` returns the string `'[]'` via `parseScalar`**
`meeting.md` frontmatter has `attendees: []` — `CompatVault.parseScalar` doesn't handle YAML arrays and returns the raw string. F12 tests don't check `attendees` so no failure here, but this is a silent trap for F14 (Tasks) or any future scenario that expects array-valued fields. A comment in `parseScalar` noting "only handles scalars — arrays/inline-objects returned as raw strings" would make the limitation visible.

**L2 — Missing-template test path uses `Templates/` prefix, real templates don't**
`tpCreateNew(vault, 'Templates/no-such-template.md', ...)` — templates are loaded as flat paths (`daily-note.md`, not `Templates/daily-note.md`), so this correctly throws, but the path choice reads like a mistake. A short comment in beforeEach would clarify.

**L3 — `asFile` helper defined here but not back-applied to `harness.test.ts`**
`harness.test.ts` still uses `file as never` casts. Not a bug — just inconsistency worth cleaning up in a future pass.

PR #227, opened 2026-05-03
test: F13 Excalidraw binary attachment compat (#124)

sotashimozono:
## Summary

Phase E F13 — Excalidraw compat scenario. Round-trips Excalidraw's mixed text + binary on-disk shape against the harness. The issue calls out **byte-equality on the binary path** as the load-bearing assertion; the 1 KB cyclic + 4 KB PNG-magic-prefixed payloads cover that explicitly.

### Hot APIs exercised
- \`vault.read(file)\` — load .excalidraw.md text
- \`vault.modify(file, content)\` — save drawing edits
- \`vault.createBinary(p, buf)\` — write attachments
- \`vault.readBinary(file)\` — round-trip attachments
- \`vault.delete(file)\` — remove an attachment

### Files
- \`fixtures/excalidraw/sample.excalidraw.md\` — minimal .excalidraw.md (frontmatter + JSON-in-markdown codeblock)
- \`excalidraw.test.ts\` — 8 tests:
  - .excalidraw.md text read + frontmatter cache
  - vault.modify round-trip with modify event
  - **1 KB cyclic-byte attachment write/read with byte-exact equality** (issue requirement)
  - 4 KB PNG-magic + xorshift32 payload (magic preserved exactly)
  - readBinary returns a defensive copy (mutation doesn't corrupt store)
  - createBinary rejects duplicate path
  - binary attachments excluded from getMarkdownFiles
  - delete + readBinary surfaces "file not found"

### Bug fix surfaced by F13
\`CompatVault.{create,createBinary,modify,delete,read,readBinary}\` used to throw synchronously inside \`Promise<X>\`-returning methods. The throw escaped Promise machinery so \`await ...rejects.toThrow\` didn't catch it. Converted to \`async\` so throws become rejections — matches the contract callers expect from a Promise-returning API. Happy paths unchanged.

### Doc
\`docs/plugin-compatibility.md\` — Excalidraw row flipped from "🟡 expected (RPC required)" to **"✅ verified-by-harness (#124 F13)"**.

### Stats
4 files changed, +211 lines / −14 lines. \`vitest run\` 51 files / 721 pass / 1 Windows-skipped.

### What's next
F14 Tasks and the nightly compat workflow remain as follow-up PRs in #124.

## Test plan
- [x] \`tsc --noEmit\` clean
- [x] \`vitest run tests/compat/\` — 34/34 (12 harness + 8 dataview + 6 templater + 8 excalidraw)

Ref #124 (F13 only)


sotashimozono (commented):
APPROVE with comments — tsc clean, 34/34 tests pass. The `async` conversion in `CompatVault.ts` is the correct fix. Binary test helpers are well-crafted.

### Medium

**M1 — `createBinary` doesn't defensively copy the input `ArrayBuffer`**
`new Uint8Array(data)` in `createBinary` creates a view sharing the caller's `ArrayBuffer`, not a copy. If the caller mutates `data` after the call, `f.bytes` changes silently — the mirror problem that `readBinary` already solves with its defensive copy. Tests pass today only because `asArrayBuffer()` uses `.slice()` which copies.

Fix:
```ts
const stored = new Uint8Array(data.byteLength);
stored.set(new Uint8Array(data));
const file = new CompatTFile(path, stored);
```

### Low

**L1 — `tags: [excalidraw]` in fixture → stored as string `'[excalidraw]'`**
Same `parseScalar` array-limitation as noted in #226. Not tested here, but reinforces that YAML arrays are not a supported harness value type yet.

**L2 — `pseudoRandomBytes` seed guard**
`seed | 0 || 0x12345678` — the `| 0` converts to int32, so a seed of `-2147483648` (truthy) won't fall back. Fine for test use, just slightly misleading as a "or seed === 0" guard.

sotashimozono:
Pushed 16b8129 addressing the review:

**Medium**
- ✅ **M1** \`createBinary\` now defensively copies the caller's ArrayBuffer (\`new Uint8Array(byteLength).set(...)\`) instead of viewing into it. New test exercises the raw shared-buffer case — \`asArrayBuffer\`'s \`.slice()\` in the existing tests would have masked the bug, so the new test bypasses it and writes to a \`Uint8Array\` that owns its underlying buffer.

**Low**
- ✅ **L2** \`pseudoRandomBytes\` seed-guard rewritten to \`(seed | 0) === 0 ? ... : (seed | 0)\`. xorshift32 actually only breaks at state=0; the \`| 0\` was int32 coercion (keeps shift ops stable), NOT a fallback filter. Comment now says so explicitly.
- ⏭️ **L1** YAML arrays-as-strings noted but no code change — limitation is documented, no test depends on the misparsed value. Worth tackling in a future "richer YAML" PR if F14 Tasks needs proper arrays.

Validation: \`tsc --noEmit\` clean, \`vitest run tests/compat/\` 35/35 (was 34, +1 for the M1 regression test), full suite no regressions.

PR #229, opened 2026-05-03
test: UI test coverage — obsidian runtime mock + OnboardingModal + SettingsTab (Phase F)

sotashimozono:
## Summary

Last open Phase F item from #126. Stand up an \`obsidian\` runtime mock so UI/settings code can be unit-tested without a real Obsidian process, then drop the \`src/ui/**\` + \`src/settings/**\` coverage excludes from \`vitest.config.ts\`.

### New
- **\`plugin/tests/__mocks__/obsidian.ts\`** — runtime shim:
  - \`Modal\`, \`Setting\` (chainable, with \`addText\` / \`addToggle\` / \`addButton\` / \`addDropdown\` components that record callbacks for test-side driving), \`Notice\` (recorded into a buffer accessible via \`recordedNotices()\` / \`clearNotices()\`)
  - \`PluginSettingTab\`, \`App\` / \`Vault\` / \`FileSystemAdapter\` / \`Plugin\` shims
  - Patches \`HTMLElement.prototype\` with Obsidian's \`createEl\` / \`createDiv\` / \`createSpan\` / \`empty\` / \`addClass\` / \`removeClass\` / \`setText\` / \`toggleClass\` helpers
  - Each \`Setting\` builds a real DOM row in its container so \`containerEl.textContent\` walks pick up rendered names + descriptions
- **\`plugin/tests/ui/OnboardingModal.test.ts\`** — 9 tests covering initial render, profile defaults, validation gates (4 separate Notice assertions), Save without testing, Skip / dismiss, and the \`alreadyDismissed\` guard
- **\`plugin/tests/settings/SettingsTab.test.ts\`** — 9 tests covering base render, Daemon section visibility (none / running / down), profile rows, Connect button, clientId saveSettings round-trip, Telemetry section, reconnect-attempts label

### Wiring
- \`vitest.config.ts\` — \`resolve.alias\` maps \`obsidian\` → the mock; environment switched \`node\` → \`jsdom\` (jsdom polyfills \`HTMLElement\` + \`document\` for the obsidian-mock to patch onto); coverage excludes drop \`src/ui/**\` + \`src/settings/**\` so the new suites count toward thresholds
- \`jsdom@^25\` added as a devDependency

### Stats
6 files changed, +1371 lines. \`vitest run\` 54 files / 752 pass / 1 Windows-skipped (was 52 / 734 — +18 new tests).

### Phase F status after this PR
- ✅ #127 large-file progress indicator
- ✅ #128 daemon health panel
[...]

PR #230, opened 2026-05-03
test: harden UI mock — fix false-positive vectors from PR #229 self-review

sotashimozono:
## Summary

Follow-up to #229 addressing the HIGH + MEDIUM findings from the post-merge self-review. The PR-229 mock + UI tests landed as scaffolding; this PR closes the false-positive vectors that were limiting the harness's actual regression value.

## Mock fixes (false-positive vectors)

- **Signature drift** — `addClass` / `removeClass` widened to `(...classes: string[])`, `toggleClass` to `(string | string[], boolean)`, `setText` to `(string | DocumentFragment)`, `createEl` / `createDiv` / `createSpan` accept the optional `callback?: (el) => void` Obsidian patterns rely on. Without these, `el.addClass('a','b')` silently dropped `'b'` and any chained-callback subtree never ran.
- **`simulate*` loud-fail** — `simulateInput` / `simulateChange` / `simulateClick` now throw if no callback was registered. A future refactor that drops `.onChange(...)` wiring will break the test loudly instead of producing a green pass.
- **Native button click** — `buttonEl.addEventListener('click', ...)` so `containerEl.querySelector('button').click()` (and the new `clickButton(root, label)` helper) actually fire the registered handler.
- New helpers: `findButton`, `clickButton`, `getSettingFor`, `getSettingsIn`, `AnyComponent` discriminated union (\`kind: 'text' | 'toggle' | 'button' | 'dropdown'\`).
- Drop `_obsidianMockPatched: boolean` from the global \`HTMLElement\` interface — moved to a `Symbol.for(...)` runtime sentinel so it doesn't leak onto every DOM element in the program.

## Test infrastructure

- `vitest.setup.ts` now auto-clears the Notice buffer before every test. Individual files no longer need a `beforeEach(clearNotices)`.
- `SettingsTab.test.ts` `fakePlugin` typed via a `SettingsTabPlugin` slice — adding a new required method to `RemoteSshPlugin` will surface as a compile error in the test rather than a silent runtime failure.
- `vi.mock` isolates the real telemetry singleton from the toggle path (no real fs.mkdir, no cross-test state).

## New behavioral tests

| Suite | Added |
|---|---|
| SettingsTab | Connect click → `openShadowVaultFor` called; Disconnect click → `disconnect` called; clientId text input drives `sanitizeClientId`; clientId blank resets to default sentinel; reconnect-attempts in-range / out-of-range; telemetry toggle change persists; daemon Restart success + failure paths |
| OnboardingModal | Browse-button empty-host guard fires \"Fill in host and username first\"; \`vi.waitFor\` replaces the fragile \`await Promise.resolve()\` x2 flush |

`internals(modal)` helper extracted in OnboardingModal tests — replaces six near-identical inline casts, so a `saveOnly` rename will fail in one place rather than six.

## Coverage delta

| Metric | Before (#229) | After |
|---|---:|---:|
| Lines | 68.78% | **69.62%** |
| Branches | 63.23% | **64.25%** |
| Functions | 64.43% | **65.29%** |
| `SettingsTab.ts` lines | 33.33% | **53.33%** |
| `SettingsTab.ts` branches | 26.66% | **60.00%** |

The big SettingsTab jump comes from the click/toggle handlers actually executing now instead of just rendering.

## Stats

5 files changed, +415 / -182. `vitest run` 54 files / 760 pass / 1 Windows-skipped (was 752 / 1).

## Test plan
- [x] \`tsc --noEmit\` clean
- [x] \`npm run lint\` clean
- [x] \`vitest run --coverage\` exits 0; thresholds met with buffer


sotashimozono (commented):
## Review: APPROVE with comments

tsc clean, 760/760 tests pass. All seven HIGH/MEDIUM findings from the #229 self-review are addressed. The most important fix — `simulate*` throwing instead of silently no-opping — is correctly implemented and makes every new behavioral test actually meaningful. `SettingsTabPlugin` slice type, `displayTab` helper, `vi.waitFor` flush, `getSettingsIn`/`settingByEl` WeakMap, and native button click wiring are all clean work.

### MEDIUM — Remaining gaps

**M1 — `OnboardingModal`: generate-key button still untested**
This was M1 in the #229 review. The `Browse…` button guard is now tested (good), but the "Generate key" button click — including the EEXIST path that fires a distinct Notice — still has zero UI-level coverage. `makeButton().simulateClick()` exists for exactly this, and `SshKeyGen` only needs a filesystem mock to stub. Cover this before the harness grows further.

**M2 — Browse button test uses querySelector instead of the new `findButton` helper**
`tests/ui/OnboardingModal.test.ts:152-155` manually does `Array.from(...querySelectorAll('button')).find(...)` — exactly the boilerplate `findButton` was introduced in this PR to eliminate. Use `await clickButton(modal.contentEl, 'Browse…')` instead; it throws if absent rather than relying on `expect(browse).toBeTruthy()` + non-null assertion.

### LOW

**L1** — `makeButton()` lines 228-236: `typeof document !== 'undefined'` branches are still dead code under jsdom (noted as L2 in #229 review). Simplify to unconditional `document.createElement('button')` and `buttonEl.addEventListener(...)`.

**L2** — `vitest.setup.ts:41-44`: no-op `afterAll` with a comment claiming it "keeps the polyfills in place". The polyfills are set at module init (lines 24-31) and never torn down regardless — the hook does nothing. Remove it.

**L3** — `vitest.setup.ts:17`: `import { clearNotices } from './tests/__mocks__/obsidian'` bypasses the `'obsidian'` alias. Import from `'obsidian'` to exercise the alias end-to-end and decouple from the physical path.

**L4** — `clickButton` does a single `await Promise.resolve()` after `b.click()`. For handlers that `await` more than once internally this is insufficient. Document the limitation in the JSDoc or drop the internal await and let callers use `vi.waitFor`.

---
*Full review artifact: `.claude/PRPs/reviews/pr-230-review.md`*

PR #231, opened 2026-05-03
feat: BRAT manifest-beta channel + release asset (Phase G)

sotashimozono:
## Summary

Phase G first checkbox from #126 — enable [BRAT](https://github.com/TfTHacker/obsidian42-brat) `--beta` installs from this repo so external testers can subscribe to the auto-update channel before the v1.0 community-store submission.

## Changes (6 files + version bump)

| File | What |
|---|---|
| `manifest-beta.json` (NEW, root) | BRAT `--beta` polls this. Tracks `manifest.json` byte-for-byte until a `next` branch is cut post-1.0. |
| `plugin/scripts/bump-version.mjs` | `npm version` now writes `manifest-beta.json` alongside the existing four files. Doc-comment spells out the BRAT contract. |
| `plugin/package.json` | `version` script `git add`s the new file. |
| `.github/workflows/version-check.yml` | New `manifest-beta == manifest` assertion guards against a hand-edit drifting the two channels. Drop this when `next` is introduced. |
| `.github/workflows/release.yml` | `manifest-beta.json` added to the GitHub Release `files:` block so BRAT can fetch it as a release asset. |
| `README.md` | Install section split into **Option A — BRAT** (recommended for testers) and **Option B — Manual**. Previous one-line BRAT mention was easy to miss; promoting it matches the Phase G bug-bash flow. |

Version bumped 0.4.125 → 0.4.126.

## Why two manifests when they're identical?

Pre-1.0 every merge IS a beta — there's no separate stable line yet. But adding `manifest-beta.json` now means:
1. BRAT testers can subscribe **today** (no infrastructure migration needed later)
2. When 1.0 stable ships, only the workflow's parity assert needs loosening; everything else stays put
3. `manifest.json` consumers (Obsidian's plugin browser fetch) keep working unchanged

## Verified locally

\`\`\`
$ node -e "for (const p of ['manifest.json','manifest-beta.json','plugin/manifest.json','plugin/package.json']) { console.log(p, '=', require('./'+p).version); }"
manifest.json = 0.4.126
manifest-beta.json = 0.4.126
plugin/manifest.json = 0.4.126
plugin/package.json = 0.4.126

$ diff manifest.json manifest-beta.json && echo OK
OK
\`\`\`

## Test plan
- [x] `npm version patch` syncs all four manifests + the new beta to the same version
- [x] `diff manifest.json manifest-beta.json` returns clean (the new CI assert will pass)
- [x] No code paths touched — pure manifest/workflow/docs change
- [ ] After merge: install BRAT in a vault, add `sotashimozono/obsidian-remote-ssh`, confirm 0.4.126 appears

Closes the `manifest-beta.json` checkbox in Phase G of #126.


sotashimozono (commented):
## Review: APPROVE

Small, well-scoped infra PR that wires the BRAT `--beta` channel into the release pipeline.

The version-check chain is sound — the new `manifest.json ↔ manifest-beta.json` parity step combines transitively with the existing `plugin/manifest.json ↔ manifest.json` assert (`.github/workflows/version-check.yml:75-80`), so all three manifest files stay byte-equal pre-1.0 with no extra plumbing. The "drop this when `next` is cut" comment is exactly the right forward-looking handoff so this guard doesn't become archaeological mystery post-1.0.

`bump-version.mjs` correctly writes `manifest-beta.json` and the `package.json` `version` script `git add`s it from the right relative path. `release.yml` adds `manifest-beta.json` to the release asset list — together with the existing `plugin/main.js` / `plugin/manifest.json` / `plugin/styles.css` entries, BRAT now has every file it needs from a release.

### LOW

**L1 — README BRAT button label hardcodes upstream UI strings**
`README.md` Option A step 2 says `click "Add Beta plugin" (not "...with frozen version")`. This depends on BRAT's exact button labels, which can drift across BRAT releases. Explicit labels are more useful for new users today, but worth knowing this is the kind of line that goes stale silently.

**L2 — Pre-existing drift `manifest-beta.json` was repairing**
The pre-PR `manifest-beta.json` was at `0.4.90` while stable was at `0.4.125` — silent drift for 35 patch versions because the bump script never wrote it. This PR's parity assert is precisely what would have caught that. Not a finding against this PR; noting it because the same shape of drift could re-emerge for any future BRAT-only file added without bump-script integration.

### Validation

| Check | Result |
|---|---|
| Type check | Skipped (manifest/workflow/docs only) |
| Tests | Skipped (no test surface affected) |
| Manual diff verification | All four version sources at `0.4.126`; root manifests byte-equal per PR description |

Ship it.

PR #232, opened 2026-05-03
test: M11 Playwright + Electron remote-→-Obsidian reflect scenarios (#125)

### LINKED ISSUE #125: M11: Playwright + Electron nightly smoke (real Obsidian, not mock)
## Context

Phase C M9 (sync-reflect E2E) and Phase E plugin compat both currently rely on mocked Vaults / FakeFileExplorer rather than a real Obsidian binary. That's by design — vitest unit + integration suites are fast and deterministic, and the M9 mock catches every regression in the production codebase the daemon → reader pipeline touches.

But the mock has known limits:

- **Plugin compat (Phase E)** can drift if a plugin reaches into Obsidian internals the FakeVault doesn't reproduce (workspace state, embedded view registries, modal lifecycle).
- **UI reflection** is asserted only as "the right `vault.trigger` event fired"; we can't catch a CSS regression that breaks File Explorer rendering against the real DOM.
- **Real Electron / Chromium quirks** (window lifecycle, IPC, DevTools-only behaviours) are completely opaque to the unit suite.

M11 closes this gap with a nightly Playwright + Electron job that drives a real headless Obsidian instance through 5 smoke scenarios.

## Proposed scope

- **Trigger**: nightly cron (separate from `bench.yml` so a flake in either doesn't block the other) + `workflow_dispatch` for spot-checks.
- **Runner**: ubuntu-latest (Linux Electron is the cheapest + most reliable).
- **Setup**: download a pinned Obsidian Linux AppImage (env: `OBSIDIAN_VERSION`), launch with `--remote-debugging-port=9222 --no-sandbox`, attach Playwright to the CDP endpoint.
- **5 scenarios** (the bare-minimum smoke set):
  1. **create** — File Explorer shows a new file when the bench writes one to the remote.
  2. **modify** — File Explorer's mtime changes after a remote modify.
  3. **delete** — File Explorer drops the path when removed remotely.
  4. **rename** — old path gone, new path present.
  5. **large-file image render** — write a 1 MB png to the remote, open it in a markdown embed, assert `` actually rendered (catches `ResourceBridge` regressions).
- **Asserts via**: Playwright `page.locator(...)` against the rendered DOM. Screenshots on failure as artifacts.

## What's NOT in scope here

- **Plugin compat scenarios** (Dataview / Templater / Excalidraw / Tasks) — those go in #124's compat harness, not in M11. Different cadence (per-plugin scripted) and different surface (plugin internals, not Obsidian core).
- **Performance assertions** — bench.yml owns latency; M11 is a correctness smoke.
- **Mobile** — manifest is `isDesktopOnly: true`; mobile is permanently won't-do (per F25).

## Cost / fragility

Major risks (in order):

1. **Obsidian binary licensing for CI** — currently downloadable without auth, but technically gray for CI use. May need to vendor a community-built Linux Electron with the open-source Obsidian view layer if upstream pushes back. Worth filing an upstream question before investing.
2. **Headless Electron + Xvfb flake rate** — historically ~1-3 % per scenario. Per-scenario retry (vitest-style `retry: 1`) controls this; a true flake is investigatable from the screenshot artifact.
3. **CI minutes** — Electron boot is ~30 s; full suite ~3-5 min. Acceptable nightly, not acceptable per-PR.
4. **Plugin install state** — first run downloads + installs each plugin we want to test. Cache the plugin tarballs in actions/cache to avoid hitting the community marketplace every nightly.

## Cross-refs

- Plan: `C:/Users/souta/.claude/plans/obsidian-remote-ssh-composed-sunset.md` §C.5 (Phase 2 of the test foundation, Option A; M11 in the M0–M12 graph).
- #124 — Phase E plugin compat (uses the lighter B-mock approach; M11 is the heavier complement).
- M9 (#105) — the unit-side E2E foundation M11 builds on.

sotashimozono:
## Summary

Closes #125 (M11). Adds 5 reflection scenarios that drive a real Obsidian Electron window through the daemon's `fs.watch` → RPC push → ChangeListener → Vault adapter → File Explorer chain. This is the path the existing mock-based unit + integration suites cannot exercise — `sync.spec.ts` covers the opposite direction (Obsidian writes → remote sees them), so the reflection direction had zero coverage.

## What was already there (and broken)

- `plugin/e2e/{playwright.config.ts, smoke.spec.ts, sync.spec.ts, demo.spec.ts, helpers/}` — full E2E scaffolding (CDP launcher, vault scaffolder, RemoteVerifier ground-truth check)
- `.github/workflows/e2e.yml` — nightly cron (03:00 UTC) + workflow_dispatch
- All of it broken: `@playwright/test` wasn't a devDep and `npm run test:e2e` wasn't a script. Workflow ran for the first time would fail at install.

## This PR

| File | Change |
|---|---|
| `plugin/package.json` (+lock) | Add `@playwright/test@^1.59` devDep + `test:e2e` / `test:e2e:reflect` scripts |
| `plugin/e2e/helpers/remote-verifier.ts` | Add `stat()` (mtime+size for the modify scenario), `rename()`, `writeBinaryFile()` (1 MB PNG fixture) |
| `plugin/e2e/reflect.spec.ts` (NEW) | 5 scenarios per #125's spec |
| `CONTRIBUTING.md` | Dev-setup snippet for running E2E locally + nightly trigger note |

### The 5 scenarios

1. **create** — remote write via SFTP → File Explorer shows the file
2. **modify** — remote overwrite → daemon-side mtime advances (UI mtime not exposed as a stable DOM attribute, so we assert the ground-truth via SFTP stat)
3. **delete** — remote unlink → File Explorer drops the path
4. **rename** — remote `rename(2)` → old path gone, new path present
5. **large-file image render** — 1 MB PNG via markdown `![[name.png]]` embed → `.naturalWidth > 0` (catches ResourceBridge regressions — a broken bridge would still render the `` tag but with `naturalWidth === 0`)

Each scenario uses a `Date.now().toString(36)` STAMP prefix so re-runs / parallel runs don't collide on remote paths. `afterAll` cleans up via `Promise.allSettled([remote.removeFile(...)])`.

## Validation

- `tsc --noEmit` clean
- `npm run lint` clean
- Vitest unit suite still green: 763 pass / 1 Windows-skipped
- `npx playwright test --list` discovers 14 tests across 4 spec files (5 reflect + 5 smoke + 3 sync + 1 demo)
- Actual run: requires Linux/macOS + Obsidian binary + Docker test sshd. Local on Windows is workflow-only (Xvfb-dependent). Nightly `e2e.yml` will run it; this PR's CI run will skip the E2E job (PR-trigger isn't on `e2e.yml`'s cron).

## Out of scope (separate issues if needed)

- Workflow's Obsidian AppImage URL points at the `nicehash/Obsidian` mirror, not the official `obsidianmd/obsidian-releases` — worth auditing before relying on the nightly result
- Upstream license clearance for using the Obsidian binary in CI (the current `e2e.yml` comment notes "if upstream objects, switch to a stub Electron shell")
- Plugin-install caching (the workflow currently re-downloads the AppImage each run unless cache hits)

## Test plan
- [x] tsc --noEmit clean
- [x] npm run lint clean
- [x] vitest 763 pass / 1 skipped
- [x] `npx playwright test --list` discovers all 5 new reflect scenarios
- [ ] After merge: `workflow_dispatch` trigger of `e2e.yml` to confirm the actual nightly path runs end-to-end


sotashimozono (commented):
## Review: APPROVE with comments

tsc clean, vitest 763/763 pass. The 5 scenarios are well-designed — STAMP prefix prevents cross-run collision, `fileLocator` uses the stable `data-path` DOM attribute rather than fragile text matching, `makeOnePngOf` builds a structurally valid 1×1 PNG seed + trailing padding (and correctly documents the Chromium trailing-bytes tolerance), and the `afterAll` cleanup uses `Promise.allSettled` so a missing file never breaks teardown.

### MEDIUM — Issues

**M1 — `afterAll` doesn't clean up `${STAMP}-rename-src.md` if test 4 fails before the rename**
`plugin/e2e/reflect.spec.ts:67-75`: the cleanup list includes `${STAMP}-renamed.md` (the post-rename destination) but not `${STAMP}-rename-src.md` (the pre-rename source). If test 4 writes the source file and then crashes before `remote.rename()` executes, the source file lingers on the remote indefinitely. Fix: add `remote.removeFile(\`${STAMP}-rename-src.md\`)` to the `Promise.allSettled` block; `removeFile` already ignores unlink errors so it's safe even when the file was renamed successfully.

**M2 — Scenario 2 (modify) asserts daemon-side mtime, not Obsidian-side reflection**
`reflect.spec.ts:94-116`: the test asserts that `remote.stat()` shows an advanced mtime after the overwrite, but does NOT assert that the changed content appeared in the Obsidian UI. The comment notes "mtime isn't exposed as a stable DOM attribute" — valid for direct mtime checks, but opening the note and asserting `.textContent` or Obsidian's file cache would confirm the reflect chain end-to-end. As-is the test verifies the SFTP write landed (trivially true) rather than that `fs.watch → RPC push → Vault adapter` delivered it. Acceptable as a first pass; flagging so it's tracked as a known gap.

### LOW

**L1 — Parameter name `r` in `waitForMtimeChange` is shadowed inside the sleep lambda**
`reflect.spec.ts:202`: `await new Promise(r => setTimeout(r, 250))` — the inner `r` (Promise resolve) shadows the outer `r: RemoteVerifier` parameter of `waitForMtimeChange`. There's no runtime bug (line 201's `r.stat` sees the outer `r` because the shadow is scoped to the arrow function), but ESLint's `no-shadow` rule should catch this. Rename the inner binding: `await new Promise(resolve => setTimeout(resolve, 250))`.

**L2 — `connectToRemote` is duplicated from `sync.spec.ts`**
`reflect.spec.ts:212-232`: the PR description notes this is intentional ("kept private to this file to avoid coupling the helpers package to a specific UI flow"). If the connection flow changes (command palette label, timing), two files need updating. Consider whether a shared `helpers/connect.ts` export is acceptable coupling given both specs already depend on the same palette flow.

**L3 — Hardcoded 10 s blanket `waitForTimeout` for connect handshake**
`reflect.spec.ts:231`: `await page.waitForTimeout(10_000)` is pre-existing (mirrors `sync.spec.ts`) but worth tracking — on slow CI hosts 10 s may be insufficient for the initial `fs.walk` to complete, causing reflect tests to race with the daemon's startup scan. A condition-based wait (e.g. `waitFor(() => fileExplorer.locator('.nav-file').count() > 0)`) would be more robust. Out of scope for this PR.

**L4 — `fileLocator` ends-with selector could match longer paths**
`reflect.spec.ts:190`: `.nav-file-title[data-path$="${fileName}"]`. If any existing vault file ends with the same suffix as a STAMP-prefixed name (extremely unlikely given `Date.now().toString(36)`), the locator would match both. The comment in the helper is clear about intent; just worth knowing the selector is `ends-with`, not `equals`.

### Validation

| Check | Result |
|---|---|
| tsc --noEmit | Pass |
| vitest run | Pass — 763 pass / 1 skipped |
| `npx playwright test --list` | 14 tests discovered (5 reflect + existing) |
| Actual E2E run | Requires Linux + Obsidian binary + Docker sshd — nightly CI only |

---
*Key gap to track: M2 (modify scenario doesn't verify Obsidian-side reflect, only SFTP write) and M1 (rename-src orphan on failure). Both are fixable as follow-ups without blocking merge.*

PR #234, opened 2026-05-03
feat: integrated remote terminal pane (xterm.js + ssh2 shell channel) (#149)

### LINKED ISSUE #149: Integrated remote terminal pane (xterm.js + ssh2 channel)
## Goal

A terminal pane inside the shadow vault, connected to the **already-open SSH session**, so the user can run shell commands on the remote without leaving Obsidian.

## Why this is on-brand

Our positioning is "VS Code Remote-SSH style". VS Code Remote-SSH ships an integrated terminal as a first-class feature — without it, the experience feels half-baked. Today the user has to alt-tab to a separate terminal app to do anything outside the vault file ops (run a script, check git, restart a service). This closes that gap.

## Architecture

\`\`\`mermaid
graph LR
    subgraph "Plugin"
        UI[ItemView with xterm.js] --> SC[SshChannel multiplexer]
    end
    subgraph "ssh2 (already connected)"
        SC --> Cli[Client.shell channel]
    end
    Cli --> Pty[Remote PTY]
\`\`\`

- **Reuse the existing SSH connection.** \`ssh2.Client\` supports multiple channels per connection — \`client.shell()\` opens a new PTY-backed stream alongside the SFTP / RPC channels. Zero connection-establishment overhead.
- **xterm.js** for the terminal widget (industry standard, ~150 KB unminified).
- **ItemView** registered via \`registerView\`, opens in the right sidebar by default (configurable).

## Scope (v1)

- Single terminal per shadow vault
- Default shell from \`SHELL\` env var; override in settings
- PTY size sync on pane resize
- Bearer-token-equivalent: terminal inherits the SSH session's auth — no extra credential prompt
- Settings: shell command, font size, scrollback lines (default 1000)

## Out of scope (v1)

- Multiple terminal tabs
- Persisting scrollback across reconnects
- Split panes
- Search-in-scrollback (xterm.js addon, can add later if asked)

## Risks

- **Bundle size**: xterm.js + xterm-addon-fit ≈ 200 KB minified → would push the bundle past the 600 KB gate. Mitigation: lazy-load (only require when terminal pane is opened) so the terminal code never enters the cold-start path.
- **PTY allocation**: requires sshd \`PermitTTY yes\` (default). Document the failure mode if disabled.
- **Resize race**: rapid resize can race xterm and the remote PTY. Debounce 100 ms.

## Effort

Medium. ~600-1000 LOC + 1 new ItemView + bundle-size investigation.

## Synergy with #131 (daemon version check)

This goes through ssh2 directly — the daemon doesn't have to know about the terminal channel. If the daemon is down for any reason, the terminal still works.

## Synergy with the \`Git integration\` issue

The git integration (separate issue) will defer pull/merge conflict resolution to "open the integrated terminal and resolve there". Without ⑥ this story is incomplete.

sotashimozono:
## Summary

Closes #149. Single-pane remote terminal in the right sidebar, backed by xterm.js and an ssh2 PTY channel multiplexed onto the already-authenticated SSH connection that `SftpClient` owns. Opening the pane adds zero connection-establishment cost — the channel rides the existing session.

## Architecture

\`\`\`
RemoteTerminalView (ItemView) → RemoteShell → SftpClient.openShell() → ssh2.client.shell()
       │                              │                                       │
       │       xterm.js + FitAddon    │     onData: chunk → term.write       │
       │       term.onData: → write   │                                       │
       └──────────────────────────────┴──────────────────────────────────────→ remote PTY
\`\`\`

## Files (10 src + 2 test + 1 workflow + 1 build)

| File | Status | What |
|---|---|---|
| \`src/ssh/SftpClient.ts\` | mod | \`openShell({rows,cols,term?,cmd?})\` mirrors existing \`openUnixStream/exec\` shape |
| \`src/ssh/RemoteShell.ts\` | NEW | Channel lifecycle wrapper: \`open/write/resize/close/isOpen\`. stdout+stderr → single \`onData\` |
| \`src/ui/RemoteTerminalView.ts\` | NEW | \`extends ItemView\`. xterm.js + FitAddon, 100ms-debounced ResizeObserver, "Not connected" + "Failed to open shell" disconnected states |
| \`src/main.ts\` | mod | \`registerView\` + \`addCommand("Open remote terminal")\` (gated on client.isAlive) + \`detachLeavesOfType\` on disconnect |
| \`src/types.ts\` + \`src/constants.ts\` | mod | Three new optional settings + defaults (font 12px, scrollback 1000) |
| \`src/settings/SettingsTab.ts\` | mod | Terminal section: shell override / font (6–32 px) / scrollback (100–100000 lines) |
| \`styles.css\` | mod | Pane styling + inlined xterm.js v5 default CSS (~3 KB, MIT preserved) |
| \`package.json\` | mod | \`@xterm/xterm@^5.5\` + \`@xterm/addon-fit@^0.10\` deps |
| \`.github/workflows/release.yml\` | mod | Bundle cap 600 → 800 KB (xterm bundles into main.js; ESM splitting unsupported in Obsidian plugins) |
| \`tests/RemoteShell.test.ts\` | NEW | 18 unit tests against fake ssh2 ClientChannel |
| \`vitest.config.ts\` | mod | \`RemoteTerminalView.ts\` added to coverage exclude (jsdom + xterm = impractical) |

## Why \`setActiveLeaf\` not \`revealLeaf\`

\`revealLeaf\` requires Obsidian v1.7.2; our manifest declares \`minAppVersion: 1.4.0\`. \`setActiveLeaf(leaf, { focus: true })\` has the same observable effect (focus + render) and ships in pre-1.0 Obsidian. Documented inline in \`openRemoteTerminal()\`.

## Why bundle xterm rather than lazy-load

esbuild code-splitting requires \`format: 'esm'\` which Obsidian plugins don't support — they require CJS. Inlined CSS for the same reason (no separate asset to ship). 770 KB main.js (was ~470 KB) is comfortably below the new 800 KB cap and well under standard Obsidian plugins (Excalidraw 5.9 MB / Templater 4.2 MB / Dataview 4.1 MB).

## Verified locally

- \`tsc --noEmit\` clean
- \`npm run lint\` clean (sentence-case + active-window-timers + no-floating-promises + minAppVersion gates all satisfied)
- \`vitest run\` 778 pass / 1 Windows-skipped (was 760, +18 from RemoteShell.test.ts)
- \`node esbuild.config.mjs production\` → 770 KB main.js (under 800 KB cap)

## Out of scope (per #149 v1)

- Multiple terminal tabs
- Persisted scrollback across re-opens
- Split panes
- Search-in-scrollback
- Back-pressure on huge pastes (ssh2 handles it; we trust \`write()\` return)

## Test plan

- [x] tsc / lint / vitest all green locally
- [x] Bundle stays under 800 KB
- [ ] After merge: connect to a real remote, run "Open remote terminal", assert keystroke roundtrip + resize + clean disconnect on \`Remote SSH: Disconnect\`


sotashimozono:
## Self-review (multi-agent: typescript / comments / silent-failures / tests)

**Decision: REQUEST CHANGES** — 5 HIGH bugs in the channel/View lifecycle surface, all confirmed by ≥ 2 independent agents. Each is < 10 LOC to fix.

### HIGH (must-fix before merge)

**1. Concurrent `RemoteShell.open()` calls leak a channel** — `plugin/src/ssh/RemoteShell.ts:44-73`
The `if (this.channel)` guard runs BEFORE the `await this.client.openShell(opts)`. Two concurrent callers both pass the guard, both allocate channels, only the second `this.channel = ch` survives — first leaks. Fix: synchronous `this.opening: boolean` flag set BEFORE the await.

**2. `close()` while `open()` is pending leaks a channel** — `plugin/src/ssh/RemoteShell.ts:46-75`
If `close()` runs while `open()`'s await hasn't resolved, `this.closed = true` is set but listeners attach + channel saved when the await resolves. No one calls `end()`. Fix: after the await in `open()`, `if (this.closed) { ch.end(); return; }` before assignment.

**3. Re-entrant `openRemoteTerminal()` opens 2 leaves** — `plugin/src/main.ts:215-218, 614-630`
`checkCallback` doesn't debounce. Rapid Cmd+P → both calls see `existing.length === 0` (await leaf.setViewState pending), both create leaves, both open shells. Fix: `private openingTerminal = false` flag.

**4. Failed `shell.open()` leaves View half-constructed** — `plugin/src/ui/RemoteTerminalView.ts:107-119`
Catch block returns without nulling `this.shell` / `this.term`. Late `'error'`/`'close'` from the partially-allocated channel calls `term?.writeln` on inconsistent state. Fix: in catch, dispose + null all three (shell, term, fit).

**5. `onClose` handler silently drops error info if `term` is null** — `plugin/src/ui/RemoteTerminalView.ts:101-104`
When teardown raced ahead, `this.term?.writeln(...)` swallows the entire error. No log, no UI banner, no audit trail. Fix: log `reason`/`cause` at warn unconditionally INSIDE the callback, before the optional-chain UI write.

### MEDIUM
6. `terminalShell` settings change requires re-open with no UX notice
7. `term.onData` listener never deregistered after channel close (silent keystroke discard)
8. `setActiveLeaf` throws are swallowed (wrap with `logger.warn`)
9. `terminalShell` is `exec`'d verbatim — local-FS injection surface (bounded threat model; document in description)

### LOW
10. Stale `#149` comment references in 5 source locations
11. Inaccurate `terminalScrollback` docstring (conflates setting with buffer content)
12. `setWindow` typed loosely
13. `renderTerminalPanel` missing explicit return type
14. Low-value WHAT comments
15. Duplicated JSDoc between `SftpClient.openShell` ↔ `RemoteShell.open`

### Test gap
- **T1**: `SftpClient.openShell()` has no unit test — 18 LOC, 1 branch
- **T2**: `RemoteShell.test.ts` misses open-then-close-while-pending race (= HIGH-2)
- **T3**: `RemoteTerminalView` excluded from coverage; testable via a `TerminalController` extraction (optional follow-up)

### Validation

| | |
|---|---|
| tsc / lint / vitest | ✅ 778 pass / 1 skipped |
| Bundle | ✅ 770 KB / 800 KB cap |
| CI | ci.yml bundle-cap fixed, commitlint body-line wrapped, rebased onto main |

Full report: `.claude/PRPs/reviews/pr-234-review.md`

---

Per the HIGH-only-cap rule from prior PRs: I'll fix the 5 HIGH bugs in this PR before merge, skip MEDIUM/LOW (tracked above for future). The 5 HIGH fixes are all narrow + uncontroversial; total ~30 LOC + 1 new test for HIGH-2.

sotashimozono:
## Re-review (post-fix at \`7d5224a\`)

**Decision: APPROVE** — all 5 HIGH bugs from the previous self-review are verified correctly fixed; only 4 LOW findings remain.

### Verified HIGH fixes

| # | Status | Notes |
|---|---|---|
| HIGH-1 (concurrent open) | ✅ | \`opening\` flag set sync before await; reset in finally on all exit paths |
| HIGH-2 (close-while-pending) | ✅ | post-await \`if (this.closed) ch.end()\` runs before channel assignment + listener attach |
| HIGH-3 (re-entrant openTerminal) | ✅ | try/finally fires on all early returns + throws |
| HIGH-4 (catch dispose) | ✅ | shell/term/fit nulled before \`renderDisconnected\` |
| HIGH-5 (onClose audit log) | ✅ | logger.info/warn unconditional; runs before optional-chain writeln |

The 2 new regression tests (HIGH-1 + HIGH-2) genuinely exercise the races they describe — \`mockImplementationOnce\` + manual \`resolveOpen\` + \`await Promise.resolve()\` correctly suspends the first \`open()\` past \`this.opening = true\`. All 18 pre-existing tests retain semantic meaning. Type safety preserved.

### LOW findings (post-fix; non-blocking per HIGH-only-cap)

1. **\`ch.end()\` race-teardown logged at \`debug_\`** — \`RemoteShell.ts:63\`. The close-while-pending race itself is noteworthy; \`warn\` would surface dangling-PTY signals that \`debug_\` hides in production.
2. **\`this.term?.dispose()\` in catch not wrapped** — \`RemoteTerminalView.ts:134\`. xterm docs say never-throws; theoretical defence-in-depth.
3. **HIGH-2 test missing \`ch.listenerCount('data') === 0\`** — \`tests/RemoteShell.test.ts:101-130\`. Current asserts can't distinguish "no listener attached" from "listener gated by \`closed\` flag." Locks in the implementation against a future reorder regression.
4. **Stale "PR #234 self-review" comment refs** — \`tests/RemoteShell.test.ts:75-79, 101-104\`. \`#234\` + \`HIGH-N\` become dead references post-merge.

### Validation

| | |
|---|---|
| tsc / lint / vitest | ✅ 780 pass / 1 skipped |
| Bundle | ✅ 770 KB / 800 KB cap |
| CI | All required checks pass (windows-latest may still be in-progress; not blocking) |

Full report: \`.claude/PRPs/reviews/pr-234-review.md\`

---

Per HIGH-only-cap: ready to merge as-is. If the author wants a clean state, all 4 LOW fixes are < 5 LOC each and can be batched into a single follow-up commit before merge.

Issue #342, opened 2026-05-11
Settings aren't persistant


SimonStnn:
### Plugin version

1.0.48

### Obsidian version

1.12.7

### OS

Windows 11

### Remote server OS / sshd version

Debian GNU/Linux 13

### Describe the bug

Updating settings on a remote isn't persistant between sessions. The settings do get written to `.obsidian/app.json`. But Obsidian doesn't load them the next launch.

### Steps to reproduce

Update a settings like `Settings > Files and links > use [[Wikilinks]]` then restart Obsidian.

### Sync log output

```text

```

sotashimozono:
﻿Thanks for the report — this is the second visible symptom of an architectural gap also affecting #341. Both stem from the shadow-vault adapter being a one-way write-through (local → remote) without a corresponding remote → local pull path on plugin startup.

Specifically for settings persistence:

- Your edit to `Files and links → use [[Wikilinks]]` is written to remote `.obsidian/app.json` via the patched adapter (so the remote IS updated — good).
- BUT the local shadow-vault disk copy of `app.json` is **not** updated by the adapter — only the remote is.
- On next Obsidian launch, Obsidian opens the shadow vault and reads `app.json` from local disk *before* the plugin has a chance to sync anything → it sees the stale version → the setting appears reverted.
- `ShadowVaultBootstrap` (`plugin/src/shadow/ShadowVaultBootstrap.ts:74`) explicitly does not pull from remote — by original design "Obsidian fills the rest on first open and leaves our pre-created files alone", which silently broke once we wanted those files to round-trip.

I've shipped a regression test framework as PR #343 that captures this. Layer 2 has 4 `it.fails(...)` cases for the shared-config allowlist: `app.json` (your exact reproducer), `appearance.json`, `core-plugins.json`, `hotkeys.json` — each asserts that what was written to remote during one session is visible to Obsidian on the next.

When the fix lands and the `.fails` markers come off, the framework guarantees this exact regression cannot ever silently reappear.

Public design analysis: `docs/en/architecture/bidirectional-sync-gap.md` in PR #343 (rendered: will appear on the docs site once merged).

Aiming to ship the actual fix in 1.0.49 — will keep this issue updated.

sotashimozono:
Sorry for making you waiting for a week!
I tried to fix this issue, so please update plugin and check it?

There still may be some issue, please make new issue.

kazink:
[...]

Issue #341, opened 2026-05-11
Sidebar isn't updated when renaming a file


SimonStnn:
### Plugin version

1.0.48

### Obsidian version

1.12.7

### OS

Windows 11

### Remote server OS / sshd version

Debian GNU/Linux 13

### Describe the bug

When renaming a file using the title field at the top of an open editor tab, the change is successfully pushed to the remote storage/filesystem, but the Obsidian internal state does not update the active editor's reference. This results in an "unlinked" state where Obsidian attempts to save or access the old filename, triggering a "File not found" error.

### Steps to reproduce

1. Open any note in a Markdown editor tab.

1. Click on the Title at the very top of the editor window.
   <img width="1335" height="241" alt="Image" src="https://github.com/user-attachments/assets/d11b1a3f-3202-4f9b-bb8f-07b1a6bb3a25" />

1. Type a new name and press Enter.

1. The file is renamed on the remote, but the editor tab remains associated with the old path.

### Sync log output

```text
command palette → "Open sync log" as described does not exist.
```

SimonStnn:
Hi @sotashimozono, great plugin! This was something I was look for for a while. I will try to fix this issue in my fork: https://github.com/SimonStnn/obsidian-remote-ssh

sotashimozono:
﻿Thanks for the detailed report! I had a careful look at the architecture and the cause is symmetric to a second issue you may not have noticed yet (#342) — both stem from the shadow-vault adapter being a one-way write-through that bypasses Obsidian's own vault-state update path.

Specifically for the title-bar rename:

- `SftpDataAdapter.rename()` (`plugin/src/adapter/SftpDataAdapter.ts:492`) sends the SSH rename + invalidates the read caches, but never fires `vault.trigger("rename", file, oldPath)`.
- Obsidian's editor tab keeps its `TFile` reference bound to the old path → next save hits "File not found".
- On RPC transport with the daemon, the `FsChangeListener` path *eventually* fires the trigger via an `fs.changed` push, but with a race window. On SFTP transport the bug is permanent — there is no recovery path.

I've shipped a regression test framework as PR #343 that captures this. Layer 1 has 5 + 4 cases (SFTP and RPC transports) all marked `it.fails(...)` — exactly your reproducer, plus a snapshot invariant that `vault.fileMap` reflects the new path after a rename.

When the fix lands and the `.fails` markers come off, the framework guarantees this exact regression cannot ever silently reappear.

[...]

PR #343, opened 2026-05-12
test(framework): 3-layer sync-test framework for #341 + #342

sotashimozono:
## Summary

Builds out the **3-layer sync-test framework** that the design review for issues #341 + #342 called for. Captures both bugs directly as failing-by-design tests (`it.fails(...)`), then adds property-based coverage, multi-transport coverage, and a public design doc.

When the #341 / #342 fixes land, removing the `.fails` markers becomes part of the fix PRs — the diff demonstrates the framework caught the regression.

### Layer 0: refactor (preparatory)

- `plugin/tests/integration/helpers/harnessVault.ts` — extracts `HarnessVault` / `HarnessTFile` / `HarnessTFolder` / `asArrayBuffer` from `sync.e2e.test.ts` so Layers 1-3 share them. No behaviour change.

### Layer 1: writer self-reflect harness — captures #341

- `helpers/assertSelfReflect.ts` — writer-side variant of `assertSyncReflect`. Same span+budget contract, FakeFileExplorer attached to the writer's own vault.
- `self-reflect.e2e.test.ts` — 5 `it.fails` cases over **SFTP transport**: write/modify/rename/delete/post-rename-fileMap.
- `self-reflect-rpc.e2e.test.ts` — same 4 cases over **RPC transport (daemon)**. Proves the bug is not SFTP-specific.

### Layer 2: restart round-trip harness — captures #342

- `helpers/assertConfigRoundTrip.ts` — seed remote `<configDir>/<basename>`, run bootstrap, compare local↔remote with order-insensitive deep-equal.
- `restart-roundtrip.e2e.test.ts` — 4 `it.fails` cases for the shared-config allowlist: `app.json` / `appearance.json` / `core-plugins.json` / `hotkeys.json`.

### Layer 3: invariant catalog + scenario runner

- `helpers/invariants.ts` — `Invariant` / `InvariantContext` / `InvariantResult` types, three catalog entries (I1: fileMap mirror, I2: matching trigger, I3: shared config round-trip), `runScenario` driver, `formatReport` diagnostic.
- `invariants.e2e.test.ts` — two `it.fails` scenarios (basic-crud, rename-chain) + always-green catalog-completeness sanity test.
- `helpers/propertyTesting.ts` — **fast-check** `opSequenceArbitrary` that produces random VALID adapter op sequences (model-tracked via in-memory `Set<string>`). When an invariant violation surfaces, fast-check automatically shrinks the failing sequence to a minimal reproducer.
- `invariants.property.e2e.test.ts` — one `it.fails` property (numRuns=5, maxOps=3) asserting I1+I2 hold for every short random sequence, plus an always-green generator self-test sampling 50 sequences.

### Public design doc

- `docs/en/architecture/bidirectional-sync-gap.md` — adapts the internal `.claude/PRPs/reviews/issues-341-342-design-review.md` for the Quartz docs site. Linked from issue comments so external readers can see the full analysis + framework map.

## Failing-by-design — CI strategy

Every regression test is marked `it.fails(...)` so vitest treats it as expected failure today. CI stays green. When the production fixes land:

- #341 fix PR drops `.fails` from 5 + 4 + 2 (SFTP self-reflect / RPC self-reflect / Layer-3 scenarios) + 1 property
- #342 fix PR drops `.fails` from 4 Layer-2 cases

That dependency makes the framework load-bearing: the fix PRs cannot merge with `.fails` still attached (an `it.fails` block that actually passes is itself a vitest failure), so the framework gates the fixes.

## Test plan

- [x] CI passes — all required checks green (`Lint & type-check`, `Test (ubuntu-latest)`, `version-check`, `SSH integration`)
- [x] All `it.fails` cases fail as expected (vitest treats them as passing)
- [x] Always-green sanity tests (catalog completeness, generator self-test) confirm wiring is correct
- [ ] After #341/#342 fix PRs, the diff drops `.fails` markers — they turn green

## Design context

Detailed analysis in `docs/en/architecture/bidirectional-sync-gap.md` (this branch) and the internal `.claude/PRPs/reviews/issues-341-342-design-review.md`.

Generated with Claude Code


sotashimozono:
failing at this time is okay.
because I only added reported in issue so it should return fail.

I need to ckeck whether I was able to implemented in an correct way...

PR #410, opened 2026-06-15
fix(shadow): restart guidance on first connect to a new profile — 1.1.4-beta.0

sotashimozono:
First beta of the 1.1.4 cycle (opened after 1.1.3 stable shipped). Fixes the **new-profile first-connect** friction surfaced while dogfooding 1.1.3.

## The bug (dogfound)

Connecting a **newly-added** profile appeared to do nothing — no vault window opened — even though the SSH connection itself succeeded. Root cause: the shadow vault is registered in Obsidian's vault list (`obsidian.json`) on first connect, but a **running** Obsidian only reads that file **at startup**, so the `obsidian://open` we fire is a no-op for a vault it hasn't loaded. A full restart (re-reads `obsidian.json`) fixes it. Confirmed from the field log: two Connect clicks 17 s apart both failed to open → not a race, it's the startup-cached vault list.

The plugin made this worse by showing a misleading **"opened in new window"** notice for a window that never appeared.

## Fix (option B — the safe, no-internal-API path)

`openShadowVaultFor` branches on `result.registryCreated`:
- **newly registered** → a **persistent notice** (duration 0): _“… is a newly registered vault. Obsidian only loads new vaults at startup … fully quit and reopen Obsidian, then click Connect again. (Only needed the first time you open a profile.)”_
- **reused** → the normal "opened in new window" notice.

This is a one-time-per-profile message; it turns a silent no-op into a clear instruction. (A seamless fix that opens the vault live without a restart would need Obsidian's undocumented vault-open IPC — tracked separately as a follow-up.)

## Docs

- `getting-started/first-connect.md`: an `[!important]` callout explaining the first-connect restart requirement; also fixed a stale shadow-vault path (`<plugin-dir>/shadow-vaults/` → `~/.obsidian-remote/vaults/`).
- `operations/troubleshooting.md`: a "Clicked Connect but no window opened (new profile)" symptom → fix entry.

## Tests / checks

`openShadowVaultFor`: newly-registered → restart notice (and NOT "opened in new window"); reused → "opened in new window" (and no restart notice). Full suite green (1122 passed / 1 skipped / 3 todo); lint + `tsc --noEmit` clean.

## Not in scope

The seamless live-open (Electron IPC to the running main process) — higher risk / undocumented API; worth a follow-up issue + probe.

### DIFF
--- a/docs/en/getting-started/first-connect.md
+++ b/docs/en/getting-started/first-connect.md
@@ -10,7 +10,7 @@ description: "What happens when you click Connect: SSH handshake, daemon auto-de
 
 obsidian-remote-ssh does not edit your remote files directly through Obsidian's vault layer. Instead it:
 
-1. Creates a **local shadow vault** under `<plugin-dir>/shadow-vaults/<profile-id>/`. This is a real Obsidian vault on your local disk.
+1. Creates a **local shadow vault** under `~/.obsidian-remote/vaults/<profile-id>/`. This is a real Obsidian vault on your local disk.
 2. Mirrors the remote vault's files into that shadow vault (lazily — large files are fetched on first open).
 3. Watches both sides for changes and syncs them through the SSH-tunneled daemon RPC.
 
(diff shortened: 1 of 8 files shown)


sotashimozono:
Mitigates #411 (does **not** close it): this PR adds the restart guidance, but the seamless live-open — opening a freshly-registered vault without a manual restart — is tracked in #411 as the real fix.

PR #450, opened 2026-07-02
fix(config+vault): per-device Obsidian config (kills the perpetual write-conflict) + hide dotfiles

sotashimozono:
Dogfooding follow-up to lazy loading (#449, merged). Two real-machine problems reported: `.obsidian` config **kept raising write-conflicts, blocking work**, and `.julia` (a huge Julia cache) cluttered + slowed the tree.

## 1. Per-device config — kills the perpetual write-conflict
**Root cause (read from the code, not guessed):** a write-conflict fires when the read cache's mtime ≠ the remote file's mtime (`SftpDataAdapter.writeBuffer`, and after every write the cache is refreshed to the fresh post-write mtime). The daemon compares mtime as `UnixMilli` on both sides (`handlers/common.go`), so a **single device's sequential note edits can't perpetually conflict**. What *can*: `app.json` / `appearance.json` / `core-plugins.json` / `hotkeys.json` were the **only** config files NOT redirected per-client — they sat at the shared identity path, so two sessions (or one device across reconnects) both wrote the **same** remote file, and every settings save tripped `PreconditionFailed` on the other's mtime.

**Fix** = make them per-device by adding the four to `DEFAULT_PRIVATE_PATTERN_BASENAMES`, so `PathMapper` redirects each into this client's `<configDir>/user/<id>/` subtree — exactly what `workspace.json` / `graph.json` / `cache` already do. No shared path → **the conflict is impossible by construction**. The existing shared-config round-trip keeps working (now on the per-client path), so each machine still gets its own remote backup + cross-session persistence. This is the user's own ask: *"config should be settable per accessing device."*

Orphaned shared copies already on the remote (`.obsidian/app.json` etc.) are simply ignored from now on; no migration needed.

## 2. Hide dotfiles by default
- **`BulkWalker`** drops dot-prefixed entries from every walk result — the full walk **and** each lazy per-folder deepen (both funnel through `walk()`), keeping only the vault config dir. Matches Obsidian's own default of hiding dot-names; a dot-DIR hides its whole subtree, not just its row.
- **`DEFAULT_WALK_IGNORE_DIRS`** gains `.julia .cargo .rustup .npm .conda .gem` so the daemon prunes these huge caches **server-side** (never walked/transferred) — the perf half of "hide `.julia` by default".

## Verification
- `tsc` OK · `lint` OK · `vitest run` OK — **1204 passed** (+ per-device `PathMapper` assertions, a `BulkWalker` dotfile-filter test). Config round-trip integration tests stay green because seed-write and pull-read both go through the same `PathMapper`. Ships **1.1.6-beta.17**.

## After merge → BRAT beta.17
- The conflict modal on `.obsidian/*` should stop entirely (each device now owns its config copy).
- `.julia` and other dot-dirs disappear from the File Explorer and are no longer walked.

**Open question for the reporter:** when the conflict modal appeared, was the file it named a `.obsidian/…` config file, or a real `.md` note? This fix eliminates the config case with certainty. If it was a note, that's a genuine two-device edit (correct 3-way-merge behaviour) — tell me and I'll dig further.

Not auto-merging — review + smoke by @sotashimozono.

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.6-beta.16",
+  "version": "1.1.6-beta.17",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/manifest.json
+++ b/plugin/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.6-beta.16",
+  "version": "1.1.6-beta.17",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/package-lock.json
+++ b/plugin/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "obsidian-remote-ssh",
-  "version": "1.1.6-beta.16",
+  "version": "1.1.6-beta.17",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "obsidian-remote-ssh",
-      "version": "1.1.6-beta.16",
+      "version": "1.1.6-beta.17",
       "license": "MIT",
       "dependencies": {
         "@xterm/addon-fit": "^0.11.0",
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "obsidian-remote-ssh",
-  "version": "1.1.6-beta.16",
+  "version": "1.1.6-beta.17",
   "description": "VS Code Remote SSH-like experience for Obsidian",
   "main": "main.js",
   "scripts": {
--- a/plugin/src/constants.ts
+++ b/plugin/src/constants.ts
@@ -16,6 +16,12 @@ export const DEFAULT_WALK_IGNORE_DIRS: readonly string[] = [
   '__pycache__', '.venv', 'venv', '.tox', '.mypy_cache', '.pytest_cache',
   'target', 'dist', 'build', '.next', '.nuxt', '.cache', '.gradle',
   '.idea', 'vendor', '.terraform',
+  // Language / toolchain caches — often enormous (a populated `.julia`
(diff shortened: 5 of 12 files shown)


sotashimozono:
## /review-pr — 6-dimension review + fixall (commit 1d75cf9)

Ran `code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`, `type-design-analyzer`, `code-simplifier`. Findings (confidence ≥ 80) and disposition:

| # | Sev | Finding | Status |
|---|-----|---------|--------|
| 1 | **CRITICAL** | `preSpawnPull` bypassed PathMapper → read the old shared path and clobbered per-device config on every shadow spawn | ✅ fixed — applies `PathMapper.toRemote` (per-client for the 4 configs, identity for community-plugins/binaries) |
| 2 | HIGH | all-dotfile remote → misleading "0 files found, check remotePath" Notice | ✅ fixed — `BulkWalkResult.hiddenCount` + "0 visible, N hidden" message |
| 3 | MEDIUM | other clients' `.obsidian/user/<id>/*` (incl the 4 new files) materialized as editable files in File Explorer | ✅ fixed — dropped the `.obsidian` filter exception (config is loaded off local disk, never the walked model) |
| 4 | MEDIUM | `CONFIG_DIR_SEGMENT` hard-coded `.obsidian`, drifts from configurable `app.vault.configDir` | ✅ fixed — exception removed, no configDir dependency; `isHiddenPath` → `startsWith('.').some()` |
| 5 | LOW | user-facing Notices/comments said "shared-config" for now-per-device files | ✅ fixed |

**Tests added/updated:** BulkWalker dotfile test (config dir now dropped) + fallback-path filter test + `hiddenCount`; PathMapper cross-machine isolation test; corrected stale `restart-roundtrip` comment. tsc/lint clean, vitest **1206 passed**.

### Deferred (follow-ups, not blocking this PR)
- **Legacy→per-client one-time settings migration** (silent-failure H1 / pr-test #3): a brand-new device that relied on pulling the *shared* config won't inherit it (its shadow is instead seeded from the source vault's config + persists per-client — no data loss for existing devices, whose local copy persists). A one-time seed from the legacy shared path would make multi-device upgrades seamless. Needs an unmapped read of the legacy path.
- **Rename `SHARED_OBSIDIAN_CONFIG_FILES` / `SharedConfigWatcher` / `push|pullSharedObsidianConfig`** — "shared" is now historical (per-device via PathMapper). Cosmetic; deferred to avoid churn across main.ts + tests.

Will file these as issues if you want them tracked separately.

PR #455, opened 2026-07-11
fix(plugin): guard shadow bootstrap against self-install when Connect is clicked inside the shadow window (1.1.7-beta.0)

bepitulaz:
## Summary

- **Bug:** clicking **Connect** (Settings profile row or the connect modal) from *inside* a shadow window re-runs `ShadowVaultBootstrap` against the very vault the plugin is running from. `installPlugin`'s rm+symlink cycle then has `src == dst` for every shared file — it deletes the real `main.js` / `manifest.json` / `styles.css` and replaces each with a symlink pointing at its own path. Obsidian can't resolve the self-referential links, so the plugin silently "disappears" from the vault on the next launch (hit this in the field: user edited a profile's transport in the shadow window's Settings, clicked Connect, restarted — plugin gone).
- **Fix 1 (UX/routing, `main.ts`):** `openShadowVaultFor()` now detects "this window already IS the shadow vault for this profile" via the `autoConnectProfileId` marker and reconnects in place (`runAutoConnect('reconnect')`) instead of re-bootstrapping + re-spawning. This also makes the Settings Connect button actually useful inside a shadow window (e.g. to apply a transport change).
- **Fix 2 (defense in depth, `ShadowVaultBootstrap.installPlugin`):** hard guard — if `sourcePluginDir` and the target `pluginDir` resolve (`fs.realpathSync`) to the same directory, skip the install and report a new `'in-place'` `pluginInstallMethod`. Protects every other path into the bootstrap, not just the button.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] CI / tooling
- [ ] Dependency update

## Test plan

- [x] Existing tests pass (`npm test`) — 80 files, 1212 passed
- [x] New tests added — regression test in `tests/ShadowVaultBootstrap.test.ts` simulating a re-bootstrap where the running plugin's dir IS the shadow's plugin dir; asserts `pluginInstallMethod === 'in-place'` and that `main.js` still resolves to the staged bundle (pre-fix this leaves a self-referential symlink and the read fails)
- [x] Manually verified against a live SSH server — reproduced the field failure on a real setup (macOS, SFTP→RPC transport switch, Connect clicked inside the shadow window → bricked install), then confirmed the fixed build connects and populates normally from the same shadow vault

## Checklist

- [x] `tsc --noEmit` passes
- [x] Bundle size within limit (`npm run build`)
- [x] No secrets or credentials committed


sotashimozono:
Thank you for contribution!

This fix seems very reasonable, great contribution.
I merged into next branch.

I will merge into main branch soon.

Issue #429, opened 2026-06-25
Claudian is not writing the files in the remote Vault but in the local folder


jsun-tech:
### Plugin version

1.1.5

### Obsidian version

1.12.7

### OS

Windows 11

### Remote server OS / sshd version

deb13.1-cloud-amd64

### Describe the bug

Hello,

first of all, thanks for your work. This is a great plugin. Actually it solves a great use case that I was struggling with for a long time.

I was trying this plugin and found the following issues:
- I install Remote SSH
- I open the remote Vault
- I Install Claudian in the remote Vault. 
- I ask Claudian (Claude Code wrapper) to create a md file.
- The file it's created in the local folder (C:/Users/.../.obisidan-remote)
- It's not reflected in the Remote (I connect through ssh separately to check)

Also, related to this issue https://github.com/sotashimozono/obsidian-remote-ssh/issues/342
- If I close the remote Vault and I open it again Claudian is not installed.

Thank you.

### Sync log output

```text

```

sotashimozono:
I tried to fix some issue and released 1.1.6.
I'm not sure this claudian issue has been fixed or not. sorry.

please let me know a result.

jsun-tech:
Hello,

thanks for making progress on this. A few comments after the update:

What works now:
- Now I see that Claudian plugin is still installed after restarting the vault

What doesn't work:
- Claudian settings are not kept after restarting the vault
- Claudian is creating still the documents (md) in the local folder in Windows 
- TaskNotes (a new plugin I am testing with Remote SSH) settings are not kept after restarting the vault
- TaskNotes seems also to automatically create a folder in local for the plugin to work.
- A error that found while trying to change the settings of a plugin (Claudian)
<img width="258" height="54" alt="Image" src="https://github.com/user-attachments/assets/f1be0948-3adf-415b-90cb-900e9a779a12" />

It seems like plugins are not aware that they should be working in the context of the remote folder.

thanks

sotashimozono:
Thank you for reporting more information.
I need to dive into more deeply...

Until it fixes, can you try with call claude from outsize and use this plugin as a native obsidian?
I will try my best but I'm not sure how can I fix it currently.

PR #459, opened 2026-07-14
fix(config): persist third-party plugin settings across restarts (#342, #429)

sotashimozono:
Fixes the "plugin settings are not kept after restarting the vault" half of #429, and #342.

## The bug

Third-party plugin settings were lost on **every** restart of the shadow vault — and the good settings on the **remote were destroyed** in the process.

## Root cause — an ordering problem a connect-time pull cannot fix

Obsidian loads community plugins during startup. Each plugin's `onload()` calls `Plugin.loadData()`, which reads `<configDir>/plugins/<id>/data.json` — **before** remote-ssh has connected over SSH and patched the adapter. That read therefore always hits the real `FileSystemAdapter`, i.e. the **local shadow disk**. We cannot get in front of it: the SSH connect is async and lands at layout-ready.

And nothing ever wrote that local copy. `saveData()` went through the patched adapter straight to the remote, so the local file stayed empty:

1. Restart → `loadData()` reads the (empty) local `data.json` → plugin boots on **DEFAULTS**
2. Plugin saves those defaults → once the adapter is patched, they're pushed to the remote
3. → **the real remote settings are overwritten** 💥

The plugin *code* survived (`main.js` / `manifest.json` / `styles.css` are round-tripped by `PLUGIN_BINARY_FILES`), which is exactly the asymmetry reported in #429: *"Now I see that Claudian plugin is still installed after restarting the vault … Claudian settings are not kept after restarting the vault."*

Note this also means `preSpawnPull` cannot be the fix: it only runs when you Connect **from the source window**. Reopening the shadow vault directly never runs it.

## The fix

**1. `SftpDataAdapter` — write-through.** A successful `<configDir>/**` write is now mirrored onto the local shadow disk with the same bytes that landed on the remote. The local disk becomes a **warm cache**, so the startup read sees the settings this device last saved, no matter how the window was launched.

- Best-effort: a mirror failure never fails the remote write that already succeeded.
- Scope is `<configDir>/**` only — the **note tree stays virtual** (served from the remote, never mirrored). Mirroring it would defeat the shadow-vault design.

**2. `PathMapper` — `plugins/*/data.json` goes per-device**, redirected into `<configDir>/user/<clientId>/`.

`saveData()` fires on *every* settings change, so a shared remote path would reintroduce the perpetual write-conflict that the four core config files were redirected to escape (b0e2773). Per-device makes a cross-device collision **impossible by construction** — two machines can never fight over one settings file.

Deliberately `plugins/*/data.json` and **not** `plugins`: the plugin's **code must stay SHARED** at the identity path, or a plugin installed on one machine would stop loading on the others. This required:
- teaching the pattern matcher a `*` segment (matches exactly one path segment, never across `/`),
- making crossing-point detection wildcard-aware, so listing a plugin's own dir merges in this client's private `data.json` (keeping `list()` consistent with `exists()`/`read()`).

## Design after this PR

| | local shadow disk | remote |
|---|---|---|
| notes | virtual (not mirrored) | source of truth |
| `<configDir>/**` | **warm cache** (write-through) | source of truth |
| plugin settings `data.json` | warm cache | **per-device** `user/<clientId>/` |
| plugin code | — | **shared** (identity path) |

## Test plan

- [x] New `PathMapper` cases: settings-private / code-shared, `*` does not over-match (`plugins/a/b/data.json`, `.obsidian/data.json`, `Notes/data.json`), per-device `toRemote`, `toVault` round-trip, listing merge at both depths.
- [x] New `SftpDataAdapter` cases: the local mirror lands, the remote gets the **per-device** path (and *not* the shared one), ordinary notes are **not** mirrored, a mirror failure does not fail the write, no-op without a shadow base path.
- [x] Full suite green — 80 files, **1223 passed**, 0 failures.
- [x] `tsc --noEmit` clean, `eslint` clean.

## Not in this PR

The other half of #429 — plugins that bypass the Vault API entirely and use Node `fs` against `adapter.basePath` (Claudian creating `.md` files in the local folder). That is a separate architectural issue, already documented as a known trade-off in `docs/en/user-guide/plugin-compatibility.md`. Tracking separately so this fix ships on its own.

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.7-beta.0",
+  "version": "1.1.7-beta.1",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/manifest.json
+++ b/plugin/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.7-beta.0",
+  "version": "1.1.7-beta.1",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/package-lock.json
+++ b/plugin/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "obsidian-remote-ssh",
-  "version": "1.1.7-beta.0",
+  "version": "1.1.7-beta.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "obsidian-remote-ssh",
-      "version": "1.1.7-beta.0",
+      "version": "1.1.7-beta.1",
       "license": "MIT",
       "dependencies": {
         "@xterm/addon-fit": "^0.11.0",
@@ -271,6 +271,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       },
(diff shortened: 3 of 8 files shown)


sotashimozono:
## Review pass — two real defects found in this PR, both fixed (64782ee)

An adversarial review of my own diff turned up two genuine bugs that the write-through had *introduced*. Both are now fixed and pinned by regression tests.

### 1. 🔴 Symlink clobber — the #455 bug class, reintroduced

`writeThroughConfig` mirrored **every** `<configDir>/**` write, and `fs.writeFileSync` **follows symlinks**. `ShadowVaultBootstrap.installPlugin` symlinks remote-ssh's own `main.js` / `manifest.json` / `styles.css` in the shadow vault back to the **source vault's real files**. So any adapter write to those paths — e.g. updating the plugin from inside the shadow window — would have written straight through the link and **overwritten the source vault's install**, bricking it. Precisely what #455 had just fixed.

**Fixed:** `lstat` the target and refuse to mirror a symlink; `realpath` the parent so a symlinked *ancestor* (a stale whole-dir plugin symlink) can't redirect the write out of the shadow root either. Skipping costs nothing — the remote write already succeeded, and for plugin **code** the local copy is owned by the pull/push binary round-trip, not by this cache.

Verified the guard is load-bearing, not decorative — with it disabled the source file is observably destroyed:

```
× never writes THROUGH a symlink (#455 regression guard)
  AssertionError: expected 'NEW-BUNDLE-BYTES' to be 'SOURCE-REAL-FILE'
```

### 2. 🔴 Path traversal

The containment guard was a `startsWith(configDir + '/')` on the **raw** string, applied *before* `..` was resolved — so `.obsidian/../evil.txt` passed it and `path.join` then resolved out of the shadow root (on win32, backslash separators were a second way through). The vault path reaches the adapter from Obsidian and from other plugins; it is not trusted input.

**Fixed:** resolve first, then verify the resolved path is the shadow root or inside it. Pinned by a test asserting nothing lands outside.

### Deliberately NOT in this PR (follow-ups)

- **`remove` / `rmdir` / `rename` are not mirrored.** Uninstalling a plugin deletes the remote copy but leaves the local shadow files, and the per-device `data.json` under `user/<clientId>/` is not cleaned up (so reinstalling the same plugin id resurrects old settings). Pre-existing for deletes — nothing was mirrored before this PR — but the per-device orphan *is* new here. Filing separately rather than widening the blast radius right before a stable cut.
- **No migration from the legacy shared remote `data.json`.** Low value in practice: the old shared copy was being overwritten with defaults on every restart by the very bug this PR fixes, so there is rarely anything worth migrating.

Full suite: **80 files, 1225 passed**, 0 failures. `tsc --noEmit` and `eslint` clean.

PR #465, opened 2026-07-14
feat(server): add extension.kill with cli.kill compatibility alias (beta.0)

lalate:
## Summary

Introduces  as the canonical lifecycle method for , with  kept as a deprecated compatibility alias.

This is **beta.0** of the incremental merge strategy proposed in [#425](https://github.com/sotashimozono/obsidian-remote-ssh/pull/425).

## Changes

- Add  struct with session-based tracking (single-principal model)
- Add  and  handlers with existence-only checks
- Register invocations in  for lifecycle tracking
- Update  to use session lookup instead of parameter passing
- Add  field to 
- Add  and  protocol types
- Fix  regex for RE2 compatibility
- Add tests for kill behavior

## Motivation

Following the incremental merge strategy from #425 review:
- **beta.0** (this PR):  +  — ready to merge
- **beta.1** (follow-up):  /  — iterate on blocking issues
- **beta.2**: next slice

Each PR stays reviewable in isolation and the git history tells a clear story of what each increment decided.

## Verification

- ✅  passes
- ✅  passes (all tests green)

### DIFF
--- a/server/config/capabilities.json
+++ b/server/config/capabilities.json
@@ -14,7 +14,7 @@
           "name": "prompt",
           "required": true,
           "maxLength": 65535,
-          "pattern": "^(?!\\s*-)[\\s\\S]+$"
+          "pattern": "^\\s*[^\\s-][\\s\\S]*$"
         }
       ]
     }
--- a/server/internal/handlers/extension.go
+++ b/server/internal/handlers/extension.go
@@ -4,8 +4,10 @@ import (
 	"bufio"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"regexp"
 	"strings"
@@ -25,6 +27,22 @@ type extensionRunner struct {
 	vaultDir string
 	seq      atomic.Int64
 	slots    chan struct{}
+
+	mu     sync.Mutex
+	active map[string]activeInvocation
+}
+
+// activeInvocation tracks a running extension process.
+//
+// The daemon uses a single-principal model — one token minted at startup
+// that every session authenticates with. *Session is a transport handle,
+// not an identity, so there is no per-session authorization boundary.
+// session is the (rebindable) stream target for live output; when nil,
+// output is persisted to the log for replay by a future reattach.
+type activeInvocation struct {
+	session    *server.Session
+	stop       func() error
+	outputMode string
 }
 
 const maxConcurrentExtensionInvocations = 4
@@ -35,6 +53,7 @@ func NewExtensionRunner(mgr *extensions.Manager, logs *extensions.LogStore, vaul
 		logs:     logs,
 		vaultDir: vaultDir,
 		slots:    make(chan struct{}, maxConcurrentExtensionInvocations),
+		active:   map[string]activeInvocation{},
 	}
 }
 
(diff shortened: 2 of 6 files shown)


sotashimozono:
Thank you for splitting this — beta.0 is exactly the shape I hoped for. The scope is clean (`extension.kill` + `cli.kill` only, no `resumeFrom` / `ReplayFrom` / `log_store.go`), CI is green, and the `capabilities.json` regex is fixed **exactly right**:

```
^\s*[^\s-][\s\S]*$
```

`[^\s-]` (not `[^-]`) is the important bit — it stops `\s*` backtracking and consuming the leading whitespace, so `"  -rf"` is rejected again. That was the subtle part and you got it.

The kill path itself I'm happy with: mutex-guarded `active` map, `os.ErrProcessDone` tolerated, unknown id → `killed:false`, `KillCompat` sharing one implementation, tests for both branches.

## 🔴 One blocking issue: beta.0 leaks processes and then wedges the runner

The `streamProcess` rewrite came across from #425 **without** the resume feature that justified it.

Before, a failed send (client gone) killed the process:

```go
if err := session.SendNotification(...); err != nil {
    _ = cmd.Process.Kill()
    return false
}
```

Now it just stops sending and lets the process run:

```go
if err := session.SendNotification(...); err != nil {
    session = nil
}
// process keeps running
```

with the comment:

> *"the process keeps running — output is already persisted above and can be replayed via `resumeInvoke` + `ReplayFrom`."*

But **`resumeInvoke` and `ReplayFrom` do not exist in this PR** — they're deferred to beta.1. Grep confirms: `resumeInvoke` appears only inside that comment. `Invoke()` has no resume branch. So in beta.0:

- the client disconnects → the process is **never killed**;
- **nothing can reattach to it** (no resume);
- **nothing reaps it** (no timeout, no reaper);
- `streamProcess` therefore never returns, so its `defer releaseSlot()` never fires.

`maxConcurrentExtensionInvocations = 4`. Four dropped sessions on long-running invocations — and long-running is precisely the case `kill` exists for — and **the extension runner is permanently wedged**: every subsequent `extension.invoke` blocks on the slot channel forever. A daemon restart is the only way out.

This is a real regression against the current `cli` behaviour, not a pre-existing bug.

### The fix — pick one, and I'd take the first

1. **Restore kill-on-send-failure for beta.0.** Keep the old semantics (`cmd.Process.Kill()` + `return false`) and move the keep-alive behaviour to **beta.1, where it lands together with `resumeInvoke`/`ReplayFrom` that make it coherent.** This keeps each increment self-consistent, which is the whole point of splitting: beta.0 should not carry half of a feature whose other half arrives later.
2. If you'd rather keep the process alive now, beta.0 needs a reaper (an idle/orphan timeout that kills and `releaseSlot`s an invocation with no session after N seconds) — but that is extra machinery for a capability nobody can use yet.

Also, while you're in there — two loose ends from the same split:

- **Dangling comment.** The `// … can be replayed via resumeInvoke + ReplayFrom` comment describes code that isn't in this branch. Reword it to whatever beta.0 actually does.
- **Dead API surface.** `ExtensionInvokeParams` gains `InvocationID` (Go + TS), and `ResumeFrom` is still declared, but `Invoke()` ignores both now that the resume branch is gone. A client sending `invocationId` gets it **silently swallowed**. Either drop those fields from beta.0 and reintroduce them in beta.1 with the handler, or reject a non-empty `invocationId` with `ErrInvalidParams("not supported yet")` so it fails loudly instead of quietly.

Fix the streaming semantics and I'll merge this into `cli` straight away — everything else here is ready.

lalate:
Addressed all feedback from your review:

### Blocking issue fixed
**Process leak & runner wedge** — restored kill-on-send-failure in `streamProcess`. When a client disconnects (send fails), the process is now killed and the slot is released via `defer releaseSlot()`. The keep-alive behaviour moves to beta.1 where `resumeInvoke` makes it coherent.

### Loose ends fixed
- **Dangling comment** — rewrote the `flush()` comment to reflect beta.0 semantics (kill on failure) instead of referencing `resumeInvoke + ReplayFrom`.
- **Dead API surface** — removed `InvocationID` and `ResumeFrom` from `ExtensionInvokeParams` (Go + TS). They're reintroduced in beta.1 alongside the handler that consumes them.
- **`killForMethod` premature unregister** — removed the immediate `r.unregisterInvocation()` call. `streamProcess`'s deferred `unregisterInvocation` now owns cleanup, so `cli.done` is sent before the entry is removed.

### Verification
`make build` ✅ · `make test` ✅ (all green)

beta.1 (PR #466) has been rebased on top of these fixes.

sotashimozono:
okay, I will check it.

server/internal/handlers/extension.go
sotashimozono, line 1: I checked. This is what I requested.

thank you for responding,

server/config/capabilities.json
sotashimozono, line 1: I also want to ask why did you change it.

sotashimozono (commented):
I reviewed codes.

I don't get why did you changed `capabilities.json`.
Anyway, it seems reasonable to merge into cli.
If there were some issue around `capabilities.json`, please fix it while it is in beta.

server/config/capabilities.json
sotashimozono: This was my review env issue, this fix was okay.
Sorry for made you confusing

PR #466, opened 2026-07-14
feat(server): implement extension.invoke resumeFrom replay (beta.1)

lalate:
## Summary

Implements  with -based reattach and  replay.

This is **beta.1** of the incremental merge strategy from [#425](https://github.com/sotashimozono/obsidian-remote-ssh/pull/425).

## Fixes 3 blocking issues from #425 review

1. **Support reattach to completed invocation via persisted log** —  now checks  instead of the in-memory  map, allowing reattach to finished invocations
2. **Fix replay vs live-stream ordering race** — Replay-then-bind ordering prevents live batches from interleaving with historical replay
3. **Persist cli.done / exit code** — Added  to persist terminal records, and  now returns  for completion status

## Changes

- : Add , ,  methods
- : Add , ,  methods
- : Add tests for new LogStore functionality
- : Add tests for resumeInvoke scenarios

## Depends on

- #465 (beta.0: extension.kill + cli.kill)

## Verification

- ✅  passes
- ✅  passes (all tests green)

### DIFF
--- a/next/proto/types.ts
+++ b/next/proto/types.ts
@@ -110,10 +110,12 @@ export interface ExtensionSchemaResult {
 }
 
 export interface ExtensionInvokeParams {
+  invocationId?: string;
   tool: string;
   args?: Record<string, string>;
   workingDir?: string;
   persist?: boolean;
+  resumeFrom?: number;
 }
 
 export interface ExtensionInvokeResult {
--- a/next/proto/types.ts
+++ b/next/proto/types.ts
@@ -62,6 +62,8 @@ export type MethodName =
   | 'server.info'
   | 'extension.schema'
   | 'extension.invoke'
+  | 'extension.kill'
+  | 'cli.kill'
   | 'fs.stat'
   | 'fs.exists'
   | 'fs.list'
@@ -108,6 +110,7 @@ export interface ExtensionSchemaResult {
 }
 
 export interface ExtensionInvokeParams {
+  invocationId?: string;
   tool: string;
   args?: Record<string, string>;
   workingDir?: string;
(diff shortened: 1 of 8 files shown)


sotashimozono:
Merged into `cli` — thanks. Findings #1 (reattach to completed via `HasLog`) and #3 (persisted `cli.done` / exit code) are solidly fixed, with tests. Good.

One thing to carry into the next slice, though — **fix #2 is inverted relative to its own comment**:

```go
// Fix #2: Replay-then-bind ordering.
// Replay persisted output BEFORE binding session ...
inv, isRunning := r.rebindInvocation(invocationID, session)   // <-- binds FIRST
...
items, done, err := r.logs.ReplayFrom(...)
sendReplay(session, ...)                                       // <-- replays AFTER
```

The code binds the session (`rebindInvocation` sets `inv.session`) *before* it replays. For a **completed** invocation `isRunning == false`, there's no live stream, so replay is clean — and that's the primary reattach case, so this is fine in practice today.

But for a **still-running** invocation, once `rebindInvocation` returns, `streamProcess`'s next flush reads `currentInvocationSession` and can push a live batch to the newly-bound session **before** `sendReplay` runs → the exact live/replay interleave this was meant to prevent. The client can reorder by `seq`, so it's not corruption, but it isn't the clean ordering the comment promises.

To actually get replay-then-bind, either replay off the log *before* calling `rebindInvocation`, or hold `r.mu` across replay+bind so no flush can slip in between. Not a `cli`-blocker (feature branch, completed-case works), just flagging so beta.2 can close it for real.

lalate:
Fixed the inverted replay-then-bind ordering you flagged:

```
// Before (bind first, replay after — gap for live interleave):
inv, isRunning := r.rebindInvocation(invocationID, session)   // binds FIRST
items, done, err := r.logs.ReplayFrom(...)                    // replays AFTER

// After (replay first, bind after — no gap):
inv, isRunning := r.lookupInvocation(invocationID)             // read-only
items, done, err := r.logs.ReplayFrom(...)                    // replays FIRST
r.rebindInvocation(invocationID, session)                     // binds AFTER
```

`lookupInvocation` reads outputMode from the active map without modifying it. `ReplayFrom` captures the log. Then `rebindInvocation` sets the session — no window for live batches to slip in between.

`make build` ✅ · `make test` ✅

PR #469, opened 2026-07-14
probe(e2e): can Node fs be routed to the remote daemon? (#429 design decision)

sotashimozono:
**Not a product change. A measurement.**

Before committing to a design for the open half of #429 (a plugin's `fs.writeFileSync` never reaches the remote), measure the renderer instead of arguing about it.

## The plan we like

Take over Node's `fs` in the renderer and route it to the remote daemon.

It's the only option that **keeps the virtual vault** — no local mirror of the note tree, which would abandon the whole shadow-vault design — and ships **no FUSE dependency**, which would kill adoption.

## The one hard problem

Patching `fs` is easy. Everything hinges on this:

```js
fs.readFileSync(p)   // SYNCHRONOUS. The remote is 10–100 ms away.
```

- **Writes have an out:** write locally, push in the background. The plugin can't tell. Claudian's case is solved by this alone.
- **Reads have none:** a synchronous call must produce bytes *before it returns*. That needs a genuine sync-over-async bridge — and whether one **exists** in Obsidian's renderer is an empirical question we've only been guessing at.

## What this probe measures, in a real Obsidian

1. **`Atomics.wait` on the renderer MAIN thread** — the ideal bridge (worker does the async I/O, main thread blocks on a `SharedArrayBuffer`, microsecond wake-up, no process spawn). **Chromium normally forbids this on the main thread.** Whether Electron permits it is *the* question, and it decides the design.
2. `SharedArrayBuffer` / `crossOriginIsolated` / `worker_threads` / `child_process` availability.
3. **Is `require('fs')` patchable** — shared module cache + writable properties. It patches, then calls through a *fresh* `require('fs')` to prove a plugin would actually see it. (Known limit it also documents: a plugin that *destructures* — `const { writeFileSync } = require('fs')` — captures the value and bypasses any patch. Nothing can fix that.)
4. **`spawnSync` round-trip latency** — the fallback bridge's real cost.
5. **Adapter read latency** — how long a sync remote read would actually freeze the UI.

It asserts only the preconditions the approach depends on, so a **red here is a finding** ("this design isn't available in the renderer"), not a product defect.

## Then what

Read the JSON it prints in CI and pick the design. If `Atomics.wait` is allowed, we get a clean sync bridge and #429 closes without touching the architecture. If it isn't, we know the real cost of `spawnSync` and can decide whether a UI freeze of *N* ms is an acceptable price for "the plugin's write actually reaches your vault" — versus today, where it silently writes to a folder nobody looks at.

This PR is a scaffold for that decision. I'd merge it only if we keep the probe as a standing environment check; otherwise it gets deleted once we've read the number.

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.4",
+  "version": "1.1.8-beta.5",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/fs-sync-bridge-probe.spec.ts
+++ b/plugin/e2e/fs-sync-bridge-probe.spec.ts
@@ -1,0 +1,202 @@
+import { test, expect } from '@playwright/test';
+import {
+  launchObsidian,
+  driveConnectFlow,
+  findShadowVaultPath,
+  waitForShadowVaultLoaded,
+  type ObsidianHandle,
+} from './helpers/obsidian';
+import { scaffoldTestVault, type ScaffoldResult } from './helpers/vault-scaffold';
+import { RemoteVerifier } from './helpers/remote-verifier';
+import { assertSshdReachable } from './helpers/sshd';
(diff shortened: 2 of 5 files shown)


sotashimozono:
Superseded by #470, which carries the probe with a commitlint-valid type. The measurements are recorded in the commit body and in #470's description.

PR #472, opened 2026-07-15
fix(server): correct replay-then-bind ordering in resumeInvoke

lalate:
## Summary

Fix for the inverted replay-then-bind ordering flagged in [#466](https://github.com/sotashimozono/obsidian-remote-ssh/pull/466#issuecomment-4977714004) (post-merge).

## Problem

`resumeInvoke` called `rebindInvocation` before `ReplayFrom`, so for a still-running invocation, `streamProcess` could push a live batch to the newly-bound session before the replay was sent — the exact interleave the ordering was meant to prevent.

## Fix

Use `lookupInvocation` (read-only) to get `outputMode`, then `ReplayFrom` to capture the log, then `rebindInvocation` to bind the session. No window for live batches to slip in between.

```
// Before: bind → replay (gap)
// After:  lookup → replay → bind (no gap)
```

## Verification

`make build` ✅ · `make test` ✅

### DIFF
--- a/server/internal/handlers/extension.go
+++ b/server/internal/handlers/extension.go
@@ -173,19 +173,23 @@ func (r *extensionRunner) resumeInvoke(ctx context.Context, p proto.ExtensionInv
 		return nil, rpc.ErrInvalidParams("extension.invoke: unknown invocationId for resume")
 	}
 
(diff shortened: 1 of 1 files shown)


sotashimozono:
it seems okay

PR #491, opened 2026-08-02
fix(adapter): materialise only on request, not on every read

sotashimozono:
`read` / `readBinary` wrote **every note they fetched** into the disk cache. That was mine to begin with, and it went beyond the stated design — a file materialises when something *asks* for it — while incurring exactly the cost that design avoids: opening a note makes a file appear inside the vault directory, where **Obsidian's own file watcher sees it arrive behind the vault model's back**.

Two E2E reds are consistent with that, and both were green before this cache existed:

```
links-metadata  getBacklinksForFile("B.md") never listed "A.md" — the forward link resolves,
                the reverse index is empty
vault-features  same, for a deep-tree pair (a/one.md ← a/b/c/deeper.md)
```

I am **not** claiming this is the cause — E2E will say — but the read-through should not have been there either way.

## After this

The disk cache is written only by an explicit request:

| request | path |
| --- | --- |
| `getFullPath` / `getFilePath` / patched `readFileSync` | `materializeSync` — bytes already in the in-memory read cache, or nothing |
| patched `fs.promises.readFile` | `materializeAsync` — one fetch, size checked against the budget from the vault model first |

`materializeAsync` moves into the adapter as the async twin of `materializeSync`, so both entry points carry the same config-tree guard, instead of `main.ts` calling `readBinary` and relying on a side effect that no longer exists.

Opening a note therefore leaves nothing on disk. The bytes are in the in-memory `ReadCache` as before, and a later `getFullPath` for that note still resolves — it writes the copy at that point.

## Notes

- Stacks version-wise with #490 (both bump to `1.1.8-beta.12` from `next`'s beta.11); whichever lands second needs a re-bump. `version-check` is not a required check, so this does not block.
- No auto-merge.

## Verification

`tsc --noEmit`, `npm run lint` (1 pre-existing warning), `vitest run` (85 files, 1312 tests) green locally. The config-tree regression test from #488 now also pins "a plain read materialises nothing".

Refs #429

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,14 @@
 name: CI
 
+# `pull_request` carries NO `branches:` filter on purpose. The gate
+# belongs wherever a PR is open, not only on the two release branches
+# — a PR into a long-lived feature track (`cli`, and any future one)
+# is exactly where an unreviewed regression can sit for weeks. Push
+# stays pinned to main/next: outside a PR there is nothing to gate.
 on:
   push:
     branches: [main, next]
   pull_request:
-    branches: [main, next]
 
 concurrency:
   group: ci-${{ github.ref }}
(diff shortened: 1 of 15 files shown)


sotashimozono:

PR #494, opened 2026-08-02
test(e2e): state the #429 read contract the product can actually keep

sotashimozono:
Stacked on #493.

## Why this touches assertions

This spec was written as a bug report — *"most of this is expected to fail"* — and it named the two possible fixes. One of them has since been **refused**, the other **taken**:

| | |
| --- | --- |
| (i) mirror the note tree onto the shadow disk | **refused** — that is a local clone of the remote vault, the thing this design exists to avoid |
| (ii) make Node's own `fs` answer for the vault | **taken** — `FsModulePatcher`, names/stats from `vault.fileMap`, content on request |

Taking (ii) surfaced a limit the original text did not anticipate: **a synchronous call cannot fetch from the remote.** Not "does not yet" — cannot. The bridge that serves vault content runs an HTTP server on the renderer's own event loop, so blocking the renderer to wait for it blocks the thread that would answer. That self-deadlock hung connect for minutes when a revision of `getFullPath` shipped it (#487), and `fs-sync-bridge-probe.spec.ts` keeps measuring it.

## What the read side now pins

```
fs.promises.readFile        → fetches on demand. That read IS the request.
fs.readFileSync             → serves only what is already here (opened, written,
getFullPath / getFilePath     or explicitly materialised). Cold → ENOENT.
fs.readdirSync / statSync   → names and sizes from the model, no download.
```

Three tests, from both directions:

1. **cold `readFileSync` → ENOENT.** This is the one assertion deliberately inverted, and it is not a weakened bug report: a green there is what proves we did **not** quietly start mirroring the vault. Its failure message says so.
2. **the same class of note IS readable via `fs.promises.readFile`** — and must produce the remote's bytes. A red here means on-demand materialisation is dead.
3. **the transition**, on a new fixture (`COLD_NOTE`) nothing else in the file touches: invisible, then visible, with nothing changed but the asking. (These tests share one Obsidian window, so "before anyone asks" needs a note of its own.)

`getFullPath` / `getFilePath` now read the note first — which a plugin resolving a TFile it is *working with* would have done anyway — and then require the path to resolve to the right bytes.

## Prose

The header and the five stale `PREDICT FAIL` banners are rewritten to match what the product does: the write side has a watcher behind it now (#481), and a listing that names the notes is a contract rather than a prediction (#485).

## Verification

`tsc --noEmit` and `npm run lint` (1 pre-existing warning) green locally; no unit-test surface. The check that matters is the E2E run on this branch — and with #492/#493 under it, its reds should finally be about the product rather than about keystrokes.

No auto-merge.

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.14",
+  "version": "1.1.8-beta.15",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/fs-visibility.spec.ts
+++ b/plugin/e2e/fs-visibility.spec.ts
@@ -31,45 +31,45 @@ import { fileURLToPath } from 'node:url';
  *
  * READ THIS BEFORE TOUCHING AN ASSERTION
  * ======================================
- * MOST OF THIS SPEC IS EXPECTED TO FAIL against the current product. That is
- * the point: it is the executable statement of the defect, and it must stay RED
- * until the product is fixed. These reds CANNOT be fixed by weakening the
- * assertions — "a file a plugin writes into the vault ends up in the vault" IS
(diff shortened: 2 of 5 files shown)


sotashimozono:

PR #499, opened 2026-08-03
test(e2e): require a receipt that connect actually ran

sotashimozono:
Stacked on #498, and acting on what #498's diagnostics reported.

## The evidence

```
plugin log (tail): {"msg":"Plugin remote-ssh v1.1.8-beta.17 loaded"}   ← the only line
visible modal buttons: <none>
```

No `fixture missing:` — so the key and the daemon are fine. The plugin loaded. And then nothing: no `promptConnect`, no handshake, no modal.

So `executeCommandById` returning `true` meant only that **Obsidian found a command and called it** — not that our plugin reacted. The fire was never a receipt.

## The change

`promptConnect`'s first act is to open `ConnectModal`, so a modal appearing within 3 s is proof the plugin reacted. `driveConnectFlow` now requires that receipt, and falls through to the keyboard path when it does not come — instead of reporting a success it cannot back up.

This also makes the remaining failure honest: if neither route produces a modal, the error is about the plugin never reacting, which is where the next round of evidence should point.

## Verification

`tsc --noEmit` green locally. No auto-merge.

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.17",
+  "version": "1.1.8-beta.18",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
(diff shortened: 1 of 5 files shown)


sotashimozono:
**#492 の付属品**なので閉じます。

中身は「発火したコマンドが実際にモーダルを出したか」を確認する 6 行のガードで、#492 の `executeCommandById` が無ければ意味を持ちません。加えてセレクタが素の `.modal` で、Obsidian 1.13 では別ウィンドウの設定画面がこれを満たしてしまうため、そのままでは誤検知します。

#492 は `next` に載せ替えて残しました。

PR #500, opened 2026-08-03
fix(e2e): drive the Obsidian window the modal is actually in

sotashimozono:
## What the evidence says

Obsidian is **not one window**, and every "connect never fired" diagnosis so far assumed it was.

Run [30773839148](https://github.com/sotashimozono/obsidian-remote-ssh/actions/runs/30773839148) settles it. The trace holds two pages in one Playwright context:

| page | size | actions | screencast frames |
|---|---|---|---|
| `page@253e…` | 1024x800 | **33** | 33 |
| `page@cfa353…` | 900x700 | **0** | 76 |

And the per-test failure screenshots pair up:

- `test-failed-1.png` — the workspace: `local_demo1-5`, status bar `Remote SSH: Disconnected`, **no modal**
- `test-failed-2.png` — **same vault** (`e2e-vault-yqL1Sc`), a window titled `Settings - … - Obsidian 1.13.4`, with **`Connect: E2E Test` open on it and a live Connect button**

So connect fired, `promptConnect` ran, ConnectModal opened — and the harness spent five rounds querying the other window. Every symptom was a true statement about a window the surface was not in:

- `driveConnectFlow` saw no receipt (#499)
- `visible modal buttons: <none>` (#498)
- smoke's `.modal-container` / `.prompt` → "element(s) not found"

## What changed

Harness only. No product change, and nothing relaxed.

- `findPageWith(page, selector)` — search **every** page in the context. Used by the connect receipt, the ConnectModal button click, and both palette drivers (`driveConnectFlow`, `runCommandViaPalette`), so each acts on the window that actually has the surface.
- `dismissBlockingModals` closes Settings via `app.setting.close()` on each window — needs no focus, unlike the Escape that was going to a window the dialog was never on. That also stops Settings owning every modal opened after it.
- Failures carry a per-window inventory (vault name, title, size, modal count, button labels, palette presence) in place of `visible modal buttons`. A window mix-up can no longer read as a dead product.
- `isVisibleWithin` replaces `isVisible({ timeout })` — Playwright documents that option as **ignored**, so the receipt check was instantaneous and raced ConnectModal's first paint even in the right window.
- Corrected: ConnectModal is one pane *or* two depending on profile count (`ConnectModal.onOpen`), not always two. The E2E fixture seeds one profile, so it renders the auth pane directly.

`smoke` 2 and 3 now assert the settings tab and palette are reachable in whichever window Obsidian puts them — which is what those tests are about. Pinning them to one page asserted a property of the harness, not the product.

## Verification

- vitest: 1310 passed; `ShadowVaultBootstrap` "falls back to the found dir" fails **only in the full parallel run** and passes standalone both with and without this change — pre-existing flake, untouched here (this PR changes `e2e/**` only, which vitest does not run)

## Not addressed here

`links-metadata.spec.ts` and `vault-features.spec.ts` drive the palette in the **shadow** window with the same single-page assumption. They have never actually run past connect, so there is no evidence yet that the split occurs there. Left alone deliberately — CI will say.

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.18",
+  "version": "1.1.8-beta.19",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/helpers/obsidian.ts
+++ b/plugin/e2e/helpers/obsidian.ts
@@ -7,6 +7,26 @@ import * as os from 'node:os';
 
 const CDP_PORT = Number(process.env.CDP_PORT ?? '9222');
 
+/** Obsidian's command-palette / Quick-Switcher text input, across versions. */
+const PALETTE_INPUT = '.prompt input, input.prompt-input, .suggestion-container input';
+
+/**
+ * Whether `locator` becomes visible within `timeoutMs`.
+ *
+ * NOT `locator.isVisible({ timeout })`: Playwright ignores that option and
+ * answers immediately, so every call site that passed one was really asking
+ * "is it visible right now" and racing whatever it was waiting for.
+ */
+async function isVisibleWithin(
+  locator: ReturnType<Page['locator']>,
+  timeoutMs: number,
+): Promise<boolean> {
+  return locator
+    .waitFor({ state: 'visible', timeout: timeoutMs })
+    .then(() => true)
+    .catch(() => false);
+}
+
 /**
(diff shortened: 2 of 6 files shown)


sotashimozono:
**必要な部分は #566 で `next` に入れた**ので閉じます。

タイトルと無関係に混ざっていた `isVisible({ timeout })` の誤用が本物でした。`isVisible()` は待たない API で `timeout` は無視されるため、「5 秒待つ」「8 秒待つ」と読める 2 箇所が実際には 1 フレームの判定でした。#566 で `waitFor` を使う `isVisibleWithin` に置き換えています。

本題の複数ウィンドウ探索（`findPageWith` / `describePages`）は現行の `obsidian.ts` と衝突します。設定画面が別ウィンドウになった件（#502 / #504 と同根）として、必要になったら現行コードに対して書き直します。

PR #501, opened 2026-08-03
fix(e2e): attach to the window that has the vault, not pages()[0]

sotashimozono:
Stacked on #500.

## Progress first

#500 worked. The Connect button got clicked and the run went further than any before it — the plugin's **own notice** now appears in the failure screenshot:

> Remote SSH: "E2E Test" is a newly registered vault. Obsidian only loads vaults at startup, so it cannot open it this session — fully quit and reopen Obsidian, then click Connect again. (One-time.)

So connect fired, `openShadowVaultFor` ran, and the shadow vault was registered. `connectShadow` already does the "fully quit and reopen" that notice asks for.

## The new wall

The specs still died, now on a 300 s `beforeAll` timeout. The trace ([run 30775609866](https://github.com/sotashimozono/obsidian-remote-ssh/actions/runs/30775609866)) is **30 seconds of polling `.workspace` on a page that does not have one** — Obsidian's **vault chooser**. Two faults put us there.

### 1. `killExistingObsidian` has never fired on Linux CI

It runs `pkill -f Obsidian` — case-sensitive — and the binary is `~/obsidian/squashfs-root/obsidian`, all lowercase (`e2e.yml:170`). So the relaunch **signalled the still-running instance** instead of replacing it, and Obsidian answered a request for a vault it had not loaded at startup with the chooser. The screenshots show the chooser and the old scaffold window alive side by side.

`-i` is **not** the fix — the checkout is at `/home/runner/work/obsidian-remote-ssh/`, so `pkill -i -f obsidian` matches the Playwright runner's own command line and kills the test run. The kill is now anchored on the resolved binary path (Electron's helpers share that argv[0]; nothing else on the box can match it) and polls for the process to actually be gone, escalating to SIGKILL, instead of sleeping 2 s and hoping.

### 2. `launchObsidian` took `pages()[0]` regardless of what it was

`selectVaultPage` picks the page whose `vault.adapter.basePath` **is** the vault we asked for; falls back to any page with a live `app.workspace` (the shadow dir is not always the path we asked to open); otherwise throws with the window inventory. It never returns a page that cannot be the vault.

`waitForObsidianReady` is deleted with it — it polled `.workspace` for 30 s and then returned anyway ("non-fatal: tests can still run even if workspace isn't visible"). They cannot: it handed back a page with no `app`, and everything after failed for reasons unrelated to the product. The selection gate throws where that shrugged.

## Verification

- `tsc --noEmit` over `e2e/**` — clean (only the pre-existing `playwright.config.ts` `actionTimeout` complaint)

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.19",
+  "version": "1.1.8-beta.20",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/helpers/obsidian.ts
+++ b/plugin/e2e/helpers/obsidian.ts
@@ -136,17 +136,28 @@ export async function findPageWith(
  * best-effort: a diagnostic that throws would replace the real error.
  */
 export async function describePages(page: Page): Promise<string> {
-  const pages = contextPages(page);
+  return describePageList(contextPages(page), page);
+}
+
+async function describePageList(pages: Page[], driven?: Page): Promise<string> {
   const lines: string[] = [];
   for (let i = 0; i < pages.length; i++) {
     const p = pages[i];
     const info = await p
       .evaluate(() => {
         const w = window as unknown as {
-          app?: { vault?: { getName?: () => string } };
+          app?: {
+            vault?: {
+              getName?: () => string;
+              adapter?: { basePath?: string };
+            };
+            workspace?: unknown;
+          };
         };
         return {
           vault: w.app?.vault?.getName?.() ?? null,
+          basePath: w.app?.vault?.adapter?.basePath ?? null,
+          hasWorkspace: Boolean(w.app?.workspace),
           title: document.title,
           size: `${window.innerWidth}x${window.innerHeight}`,
           modals: document.querySelectorAll('.modal-container').length,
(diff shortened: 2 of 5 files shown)


sotashimozono:
Scoping what the evidence actually supports, so the causal story is not overclaimed.

**Certain:**
- `.workspace` was never visible for 30 s on the page `launchObsidian` attached to (trace: 60 consecutive `isVisible('.workspace')` polls, all false).
- `test-failed-1.png` for `fs-visibility` is Obsidian's **vault chooser** ("Create new vault / Open folder as vault", Version 1.8.9) — a window with no `.workspace`.
- `pkill -f Obsidian` cannot match `~/obsidian/squashfs-root/obsidian`. That guard has never fired on Linux CI.
- The plugin's "newly registered vault" notice was on screen, so connect ran to `openShadowVaultFor`.

**Inference, with an alternative:** I attributed the chooser to the spawn signalling a still-running instance. It could equally be Obsidian answering the `obsidian://open` that `WindowSpawner` fires for a vault it did not load at startup — which is exactly what the product's own notice describes. I cannot separate the two from this run.

Both changes here are correct under either mechanism: `selectVaultPage` refuses to drive a page that is not a vault regardless of why the chooser is up, and the kill fix is a real defect independent of it.

**Not usable as evidence:** the trace ending at 38 s. These specs do not use Playwright's `context` fixture — `launchObsidian` opens its own `connectOverCDP` browser — so a second launch lands in a context the trace never covered. The gap is a tracing artefact, not a hang.

sotashimozono:
**核心は #566 で `next` に入れた**ので閉じます。

この PR が見つけていた `pkill -f Obsidian` の不一致は本物でした。CI の実体は `~/obsidian/squashfs-root/obsidian`（全部小文字、`e2e.yml:170`）なので、**単一インスタンス保護は Linux CI で一度も発火していません**。

ただし小文字にするだけでは逆に危険で、CI のチェックアウトが `.../obsidian-remote-ssh/...` 配下のため `-f obsidian` はテストプロセス自身に当たります。#566 では `resolveObsidianPath()` が返す実バイナリを対象にし、消えるまでポーリングする形にしました。

残りの `pages()[0]` 問題（複数ウィンドウ探索）は現行の `obsidian.ts` と衝突するため、この PR としては取り込みません。必要になったら現行コードに対して書き直します。

PR #502, opened 2026-08-03
fix(e2e): the Settings window has no .modal-container

sotashimozono:
Stacked on #501.

## The inventory from #500 paid for itself on its first run

From `E2E (core)` on [run 30775609866](https://github.com/sotashimozono/obsidian-remote-ssh/actions/runs/30775609866):

```
#0 (driven) vault=e2e-vault-PEhgGR  "New tab - … - Obsidian 1.13.4"   1024x800  modals:0
#1          vault=null              "Settings - … - Obsidian 1.13.4"   900x700  modals:0
                                    modalButtons:["Turn on and reload","Browse","Check for updates"]
```

Window #1 reports **`modals:0`** while `.modal button` on that same page returns three settings buttons. So the standalone Settings window has a `.modal` but **no `.modal-container`** — and `.modal-container` is exactly what every sweep and assertion in this file anchored on.

That single fact explains both remaining symptoms:

- `dismissBlockingModals` called the workspace clean **about the one window Settings was open in**, so it never closed it.
- smoke 2 reported "Settings did not open in any Obsidian window" — while printing an inventory showing it open.

## Changes

- **`SETTINGS_ROOT`** — anchor Settings on the tab rail, not `.modal-container`. smoke 2 now asserts the **Remote SSH tab** itself: that is what the test claims, and the tab is its own evidence.
- The sweep falls back to **Escape on that page** when `app.setting.close()` has no `app` to call — window #1's line reads `vault:null`, so it may not have one. A CDP key event reaches the target page without needing OS focus.
- **`CONNECT_MODAL`** — the receipt in `driveConnectFlow` was worse than useless as a bare `.modal`: the Settings window carries one, so a window ConnectModal was never in could satisfy the receipt, and the click that followed would find no button and report nothing wrong. Now it matches ConnectModal only (auth pane always renders `.remote-ssh-connect-button`; the multi-profile picker renders the "Connect to remote vault" heading).
- The inventory carries `bareModals` and `settingsTabs` alongside `modals`, so the next red says which shape it found rather than leaving the two indistinguishable.

## Already green

`smoke 3 — command palette shows Remote SSH commands` **passed** on #500's core run (714 ms). That half of the window split is fixed.

## Verification

- `tsc --noEmit` over `e2e/**` — clean (only the pre-existing `playwright.config.ts` `actionTimeout` complaint)

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.20",
+  "version": "1.1.8-beta.21",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/helpers/obsidian.ts
+++ b/plugin/e2e/helpers/obsidian.ts
@@ -11,6 +11,32 @@ const CDP_PORT = Number(process.env.CDP_PORT ?? '9222');
 const PALETTE_INPUT = '.prompt input, input.prompt-input, .suggestion-container input';
 
 /**
+ * Obsidian's Settings UI, in either shape it takes.
+ *
+ * In its own window (1.13.4 under Xvfb) Settings is NOT wrapped in a
+ * `.modal-container` — run 30775609866's inventory reads
+ * `modals:0` while `.modal button` still returns
+ * `["Turn on and reload","Browse","Check for updates"]`. So `.modal-container`,
+ * which every sweep and assertion here anchored on, matches nothing in the
+ * window Settings is actually in: `dismissBlockingModals` called the workspace
+ * clean and smoke 2 reported "Settings did not open in any Obsidian window"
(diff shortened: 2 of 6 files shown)


sotashimozono:
**現行コードと衝突し、単体では成立しない**ので閉じます。

指摘自体は今も有効です。Obsidian 1.13 の設定画面は独立したウィンドウで、`.modal-container` を持ちません。`next` の `describeOpenModals` は今も `document.querySelectorAll('.modal-container')` だけを見ており（`obsidian.ts:573`）、結果として `dismissBlockingModals` は 0 件を返して「掃除済み」と報告する一方、別ウィンドウの設定画面は開いたまま以降のモーダルを奪います。

`smoke.spec.ts` は `next` で既に `app.setting.pluginTabs` を読む形に変わっていて事実を認めていますが、**ヘルパー側は手つかず**です。

この PR は #504 と実質 1 つの修正が 2 コミットに分かれたもので、どちらも現行の `obsidian.ts` と衝突します。必要になったら現行コードに対して書き直す方が速いと判断しました。

PR #495, opened 2026-08-02
test(e2e): click through both ConnectModal panes, and fix COLD_BODY

sotashimozono:
Stacked on #494.

## The one that matters: ConnectModal has two panes

`ConnectModal` shows a **profile list**, each row with its own Connect button (`ConnectModal.ts:54`). Choosing one opens **that profile's pane**, whose `.remote-ssh-connect-button` (`ConnectModal.ts:115`) is what actually fires the handshake.

The harness clicked once and stopped — leaving the modal parked on the second pane. No shadow vault appeared, and every spec reported:

```
findShadowVaultPath: no shadow vault entry appeared — connect command likely never fired
```

**while the command had fired perfectly.** That message has been misleading us for several runs. #492 (registry instead of keyboard) and #493 (wait for the command to be registered) are what removed the other two candidates and left this as the only one standing — the failure text stayed identical while its cause changed underneath.

Now both panes are clicked, preferring the pane's own button once it is up. A profile that needs no second confirmation just has nothing left to click, so absence stays non-fatal.

## Two supporting fixes

- **Nine specs open-coded the fragile connect.** `driveConnectFlow` + a bare `findShadowVaultPath(…, 15_000)` — one attempt, no retry — while `connectAndWaitForShadowVault`, which retries, sat unused. They now use it. Its budget goes 45s → 120s (30s per attempt): a cold CI connect pays for plugin load, daemon spawn and the first SFTP handshake, and a fast connect still returns the moment the entry appears.
- **`COLD_BODY` was referenced but never declared** — my slip in #494, which the last run caught as `ReferenceError: COLD_BODY is not defined`. The check meant to catch it matched the *seeding* line, which names the same constant, so it reported success while the declaration was missing.

## Verification

`tsc --noEmit` and `npm run lint` (1 pre-existing warning) green locally. The check that matters is the E2E run on this branch — this is the first one where a red should mean the product.

No auto-merge.

Refs #429

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #503, opened 2026-08-03
fix(e2e): never click a Connect button that is already connecting

sotashimozono:
Stacked on #502.

Two forms of one defect: **a click that cannot land was allowed to cost the whole test.**

## 1. The second click on a disabled button

`ConnectModal.renderConnectButton`'s handler sets `btn.disabled = true` and relabels the button `Connecting…` as soon as the handshake starts. The two-stage click loop then re-found that same `.remote-ssh-connect-button`, saw it was still *visible*, and clicked it again. Playwright's actionability check waits for a disabled button to become enabled — and this one never does — so with no explicit timeout the call ate the remaining budget.

That is exactly what [run 30775609866](https://github.com/sotashimozono/obsidian-remote-ssh/actions/runs/30775609866) reports:

| spec | budget | reported at |
|---|---|---|
| `smoke 4 — connect to remote vault via command palette` | 120 s exceeded | `helpers/obsidian.ts:415` — the `waitForTimeout(400)` right after the click |
| `connect-lifecycle` | 240 s exceeded | `helpers/obsidian.ts:446` — the retry sleep in `connectAndWaitForShadowVault` |

Which made the harness look like a product hang.

`clickConnect` replaces the loop. The pane count is a property of the profile list, not something to probe twice (`ConnectModal.onOpen`): one profile renders the auth pane directly, several render a picker whose per-row Connect opens that pane. `clickIfEnabled` checks `isDisabled()` first and passes an explicit timeout, so a button that cannot be clicked costs seconds.

## 2. The general form — `actionTimeout` was in the wrong place

```ts
timeout: 120_000,
// Playwright's DEFAULT actionTimeout is 0 — wait FOREVER. […] a bare `.click()`
// blocked until the whole test timed out: three minutes burned with NO message,
// which reads as "the product hung" when in fact the harness was waiting.
actionTimeout: 30_000,   // ← top level, where this is not an option
retries: 1,
```

`actionTimeout` is a **`use`** option. At the config's top level it is not an option at all — `tsc` says so outright:

```
error TS2769: … 'actionTimeout' does not exist in type 'Config<PlaywrightTestOptions …>'
```

So the bound its own comment describes had **never once applied**, and every unbounded click in the suite could still take a whole budget. Moved into `use`; the `e2e/**` typecheck is now completely clean, which is the confirmation it is in a valid position.

## Checked and left alone

`.github/workflows/e2e.yml:44` lists `plugin/playwright.config.ts`, which does not exist — the file is at `plugin/e2e/playwright.config.ts`. It is **not** a coverage gap: `plugin/e2e/**` on the line above already covers it. Redundant, not broken, so out of scope here.

## Verification

- `tsc --noEmit` over `e2e/**` — **clean, no errors at all** (previously had the `actionTimeout` one)

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.21",
+  "version": "1.1.8-beta.22",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/helpers/obsidian.ts
+++ b/plugin/e2e/helpers/obsidian.ts
@@ -436,26 +436,62 @@ export async function driveConnectFlow(page: Page): Promise<void> {
   }
 
   // Click Connect, in whichever window ConnectModal rendered in.
-  //
-  // The modal is one pane or two depending on the profile count
-  // (`ConnectModal.onOpen`): a single profile renders the auth pane straight
-  // away, several render a picker first whose per-row Connect opens that
-  // profile's pane. Either way `.remote-ssh-connect-button` is the button that
-  // fires the handshake, so prefer it and fall back to any Connect-labelled
-  // button for the picker row. A pane with nothing left to click is not an
-  // error — that is the shape of a profile needing no confirmation.
-  const clickPage = flowPage;
-  if (clickPage !== null) {
-    for (let stage = 0; stage < 2; stage++) {
-      const paneBtn = clickPage.locator('.modal .remote-ssh-connect-button').first();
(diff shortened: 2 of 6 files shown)


sotashimozono:
**必要な半分は #566 で `next` に入れた**ので閉じます。

この PR の本題（`clickConnect` の二重クリック回避）は、#495 自身が持ち込んだループへの対処で、現行の `next` は 1 回しかクリックしないため不要になっていました。

一方、タイトルと無関係に混ざっていた `playwright.config.ts` の修正が**本物の不具合**でした。`actionTimeout: 30_000` がトップレベル（41 行目）にあり、`use:` は 111 行目。`actionTimeout` は `use` オプションなので **Playwright に一度も読まれておらず**、上のコメントが謳う 30 秒上限は存在しませんでした。これは #566 で `use` の中へ移しています。

PR #504, opened 2026-08-03
fix(e2e): close the Settings window for real, and report it truthfully

sotashimozono:
Stacked on #503.

## #503 turned the suite over

| shard | #503 | before |
|---|---|---|
| **features** | **✅ 27 passed (36.3 s)** | 27 m 37 s, total loss |
| stress | 18 passed / 3 failed (2.2 m) | 25 m 36 s, total loss |
| core | 16 passed / 3 failed / 3 flaky (15.1 m) | 1 h 11 m, total loss |

The old wall-times were almost entirely timeout burn, not work.

Two of the three core failures are the harness, and **both are mine**.

## 1. smoke 2 — the tab is found, the dialog will not close

#502's half works: `.vertical-tab-nav-item[data-setting-id="remote-ssh"]` resolves. The failure is the teardown I added — Escape does **not** close the standalone Settings window:

```
Expected: hidden / Received: visible / Timeout: 5000ms
14 × locator resolved to <div data-setting-id="remote-ssh" class="vertical-tab-nav-item tappable">…
```

`app.setting.close()` is the product's own path, but it needs an `app`, and that window's inventory line reads `vault:null`.

`closeSettings` tries all three: the API on every window, Escape on the window holding the dialog, and finally **closing that window**. It is a stray Obsidian surface the harness itself opened by clicking "Trust author and enable plugins" — not a product assertion — and leaving it up has already cost this suite days. The driven page is never closed. smoke 2 asserts the result, because a Settings window nothing can close is inherited by every later spec.

**And it now returns the truth.** The version I wrote in #502 logged `closed` after merely *trying*, so a dialog nobody could close reported as one that had been. That is the same silent-failure shape this whole stack has been unpicking, and I put it in.

## 2. `fs-sync-bridge-probe` never had a hook budget

Its `beforeAll` does two Obsidian launches, a real SSH connect and a full shadow-vault load, with no `test.setTimeout`. That never fit in the default 120 s — it only *looked* like it did because the hook used to fail in ~1 ms, before connect worked at all. Now it reports:

```
Test timeout of 120000ms exceeded.
Fixture "trace recording" timeout of 120000ms exceeded during setup.
```

Given the same 300 s `images.spec.ts` gives the same shape of work, for the reason its own comment states. **No assertion is touched.**

## Left alone: the third failure is a product signal

`connect-lifecycle (SFTP)` fails with the plugin's own log line:

```
[error] Connect failed: Connection timed out  (original: "Connection timed out after 30000ms")
```

30 s to the millisecond after `installMissingShadowPlugins`, in the shadow vault's auto-connect. The three flaky specs (`connect-reconnect`, `plugin-code-roundtrip`, `restart-settings`) show the same line and pass on retry, so it is intermittent. Not touched here — it needs its own investigation, and it is exactly the kind of red this stack was built to expose.

`fs-visibility`'s remaining 3 in the stress shard are likewise product-side (#429 `getFullPath` / `getFilePath` materialisation), not harness.

## Verification

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.22",
+  "version": "1.1.8-beta.23",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/fs-sync-bridge-probe.spec.ts
+++ b/plugin/e2e/fs-sync-bridge-probe.spec.ts
@@ -52,6 +52,14 @@ let remote: RemoteVerifier;
 let shadowVaultPath: string;
 
 test.beforeAll(async () => {
+  // Two Obsidian launches, a real SSH connect and a full shadow-vault load.
+  // The default 120 s budget never covered that; it only looked like it did
+  // because the hook used to fail in ~1 ms, before connect worked at all. Now
+  // that it runs, run 30778480180 reports "Test timeout of 120000ms exceeded /
+  // Fixture 'trace recording' timeout … during setup". Same budget
+  // `images.spec.ts` gives the same shape of work, for the same reason.
+  test.setTimeout(300_000);
+
   await assertSshdReachable();
   remote = new RemoteVerifier();
   if (!(await remote.connect())) throw new Error('RemoteVerifier could not connect to the test sshd');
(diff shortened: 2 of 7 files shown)


sotashimozono:
**#502 と一体で、現行コードと衝突する**ので閉じます。

`closeSettings` は `next` に存在せず、`dismissBlockingModals` は駆動中のページに `Escape` を送るだけ（`obsidian.ts:539`）で、対象セレクタが別ウィンドウの設定画面に当たりません。3 段構え（`app.setting.close()` → Escape → 残ったウィンドウを閉じる）という発想は妥当です。

ただし #502 と同じ根に対する 2 コミット目で、単体では成立しません。また `fs-sync-bridge-probe.spec.ts` への `test.setTimeout` 追加は、そのファイル自体が `next` に存在しないため無効です。

設定画面が別ウィンドウである件は #502 のコメントに残しました。

PR #505, opened 2026-08-03
test(e2e): verify #490 against a working E2E

sotashimozono:
Stacked on #504. **This one is the product fix — it merges #490 in so the suite can adjudicate it.**

## The remaining stress failures are #490's bug, verbatim

#490's commit message, third bullet:

> `materializeSync` sees `existsSync` say yes and reports a file materialised when nothing is there, so **`getFullPath` hands back a path to nothing while claiming success**.

[Run 30779545596](https://github.com/sotashimozono/obsidian-remote-ssh/actions/runs/30779545596), stress shard:

```
getFullPath('remote_demo1.md') → /home/runner/.obsidian-remote/vaults/E2E Test--vault (5)/remote_demo1.md
<<unreadable: Error: ENOENT: no such file or directory, open '…/remote_demo1.md'>>
```

The mechanism is visible in the source. `materializeSync` short-circuits on

```ts
// Already real on disk — materialised earlier, or written by somebody. One stat, no network.
if (this.shadowBasePath && fs.existsSync(nodePath.join(this.shadowBasePath, normalizedPath))) {
  return true;   // ← returns success without writing anything
}
```

and that `fs` is the **shared** module, whose `existsSync` this plugin's own `FsModulePatcher` has replaced with one that answers from `vault.fileMap`. So it says yes for every note in the model, `materializeSync` reports success, `getFullPath` returns the path, and the patched `readFileSync` — reaching the same short-circuit — falls through to the real filesystem and ENOENTs.

That is precisely the observed pair: `existsSync` true, `readFileSync` ENOENT, nothing on disk.

## Why it was never caught

`fix/native-fs-snapshot` (#490) is a **sibling** of this stack off `next`, not an ancestor. No E2E run in this whole sequence has included it. It was also written while every connect-dependent spec died before reaching any product code, so it has never been verified by anything but unit tests.

## What is in this PR

The merge, and a version bump. **No code of mine on top.** If the two `fs-visibility` failures go green, #490 is the fix; if they do not, the cause is elsewhere and we will have learned that without guessing.

## Verification so far

- `tsc --noEmit` (src) — clean; `e2e/**` — clean
- vitest — **1315 passed**, 86/86 files (including `nativeFs.test.ts` from #490; the `ShadowVaultBootstrap` parallel-run flake was green this time)

## Still open, not addressed here

`fs-sync-bridge-probe` (core) still reports `Test timeout of 120000ms exceeded / Fixture "trace recording" timeout … during setup` even with the 300 s hook budget #504 gave it — so the time is going somewhere other than the hook. That spec probes whether a **synchronous** XHR can reach the daemon, which is the self-deadlock the design line forbids; it needs its own look rather than a bigger number.

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.23",
+  "version": "1.1.8-beta.24",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.11",
+  "version": "1.1.8-beta.12",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/manifest.json
+++ b/plugin/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.23",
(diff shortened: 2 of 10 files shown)


sotashimozono:

PR #506, opened 2026-08-03
test(e2e): make the #429 reds name the materialisation fault

sotashimozono:
Stacked on #505.

## Where #505 left us

**#490 fixed what it claimed** — `getFullPath` and `getFilePath` are gone from the failure list. Three others took their place, and they split cleanly in two:

| | failure | symptom |
|---|---|---|
| **A** | `fs-visibility:493`, `:576` | a note nobody asked for reads back synchronously — `read.error` is `null` where the contract wants `ENOENT` |
| **B** | `fs-visibility:770`, `plugin-code-roundtrip:344` | a plugin's `writeFileSync` under `basePath` **succeeds locally and never reaches the remote** |

## There is an obvious story, and that is the problem

#490's own commit message says the patched `existsSync` made `materializeSync` *"report a file materialised when nothing is there"* — so before it, nothing was ever written and a cold read was `ENOENT` **for the wrong reason**. It also says the patched `readdirSync` made the watcher's rescan *"push the whole tree back to the remote"* — which would have carried the plugin's file with it.

If both hold, #490 removed a bug that was **accidentally satisfying B**, and made materialisation real enough to **expose A**.

That is a story, not evidence. This suite has already cost five rounds of confident wrong causes, and nothing in CI can currently settle it: these specs never attach the shadow vault's own log, so *"the watcher never saw the file"* and *"the watcher pushed and the remote rejected it"* look identical from the outside.

## What this adds

`dumpMaterialisationState` reports the three things that separate them:

1. **what is really on the shadow disk** — taken from the *test runner's* `fs`, which is the real one. `FsModulePatcher` only replaces the module inside Obsidian's renderer, so this is the one vantage point that cannot be told a virtual story.
2. **the materialisation ledger** — `adapterMgr.materialized` entry count and bytes, i.e. what the plugin believes it put there.
3. **the product's own log lines** — `LocalWriteWatcher: watching the shadow vault root …`, `LocalWriteWatcher: pushed out-of-band write …`, `MaterializedCache: … exceeds the cache limit`. A push that never happened becomes distinguishable from one that was refused.

For `:770` it is attached to the **failure**, not to `expect.poll`'s `message` — that message is built before the wait starts, and the question is what happened during it.

**No assertion, matcher or timeout is changed.** One CI round should name both faults.

## Verification

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.24",
+  "version": "1.1.8-beta.25",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/fs-visibility.spec.ts
+++ b/plugin/e2e/fs-visibility.spec.ts
@@ -3,6 +3,7 @@ import type { Page } from '@playwright/test';
 import {
   launchObsidian,
   connectAndWaitForShadowVault,
+  dumpMaterialisationState,
   findShadowVaultPath,
   waitForShadowVaultLoaded,
   type ObsidianHandle,
@@ -524,7 +525,8 @@ test.describe('a plugin\'s view of the filesystem must be the REMOTE vault (#429
       'a synchronous read of a note nobody has asked for returned bytes. That means either ' +
       'the vault is being mirrored to disk — the design this plugin exists to avoid — or ' +
       'a blocking fetch crept back in, which deadlocks: the bridge that would serve the ' +
-      'content runs on the very thread being blocked.',
+      'content runs on the very thread being blocked.\n' +
+      `  ${await dumpMaterialisationState(obsidian.page, shadowVaultPath)}`,
     ).toContain('ENOENT');
   });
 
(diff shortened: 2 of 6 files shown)


sotashimozono:
Correcting the attribution in the description before it misleads anyone — I checked each of these against #504 (the last run without #490) rather than assuming.

| test | #504 (no #490) | #505 (with #490) | attributable to #490? |
|---|---|---|---|
| `fs-visibility:631` `getFullPath` | ✘ failed | ✓ gone | **fixed by it** |
| `fs-visibility:696` `getFilePath` | ✘ failed | ✓ gone | **fixed by it** |
| `fs-visibility:493` cold sync read | **✓ passed (28 ms)** | ✘ failed | **yes — regression** |
| `fs-visibility:576` readFileSync … and not before | **✓ passed (30 ms)** | ✘ failed | **yes — regression** |
| `plugin-code-roundtrip:344` `push:` | **✓ passed (3.8 s)** | ✘ failed | **yes — regression** |
| `fs-visibility:770` REAL-PLUGIN | **`-` did not run** | ✘ failed | **no** — it has never run in this stack |

So #490 is **2 fixed, 3 regressed**, and `:770` is a long-standing #429 red that is simply getting to run for the first time. My description put `:770` in cluster B as if #490 caused it; it did not.

What this does sharpen: `plugin-code-roundtrip:344` is a **push** test that passed before #490 and fails after. That is real support for the "the patched-`readdirSync` write storm was carrying pushes that the legitimate path does not" reading — but it is still a reading, and the log lines this PR adds are what would settle it.

Reading `LocalWriteWatcher.flushOne` while waiting, the gates a plugin-written file has to clear are: `pushed` signature match → `isMirroredCopy` → **two-sample stabilisation** (first sighting always defers a debounce) → size limit → push. Any of the first three would produce silence rather than a failure, which is exactly why the log tail matters.

sotashimozono:

PR #507, opened 2026-08-03
fix(watcher): scan what is already on disk when the write-back arms

sotashimozono:
Stacked on #506. **Product fix.** #506's diagnostics named the fault in one round.

## The evidence

From `fs-visibility:773` on [run 30782633453](https://github.com/sotashimozono/obsidian-remote-ssh/actions/runs/30782633453):

```
on the real shadow disk (2 files): e2e-fs-mscoxhzg-claudian.md (67b), remote_demo1.md (198b)
materialisation ledger: {"entries":0,"bytes":0}
plugin log (watcher/cache):
  LocalWriteWatcher: watching the shadow vault root for out-of-band writes
  LocalWriteWatcher: watching the shadow vault root for out-of-band writes
```

The watcher is armed. The plugin's 67-byte note is sitting right there. And there is **not one `pushed out-of-band write` line** in the whole log.

## The fault

`start()` only registered a `watch` callback, and `watch` reports **changes**. A file already on disk when we arm produces no event, ever — so it was invisible to the write-back for the entire session.

The case that matters most is exactly that shape: **a plugin writing under `basePath` in its own `onload()`**, which runs while the connect is still in flight. That is the field report this feature exists for — *"Claudian creates .md files in the local folder instead of the remote vault"*.

It *looked* like it worked before, because `scan()` read the **patched** `readdirSync`: one filename-less platform event enumerated the whole virtual vault and swept such files up with it. That is the write storm #490 removed — and removing it revealed the legitimate path had never covered them.

| test | #504 (no #490) | #505 (with #490) | |
|---|---|---|---|
| `plugin-code-roundtrip:344 push:` | ✓ 3.8 s | ✘ | real regression — the storm was carrying it |
| `fs-visibility:773` | `-` did not run | ✘ | **not** a regression — this bug, running for the first time |

## The fix, and the judgement call in it

`start()` now seeds `dirty` from the same `scan()`.

Those paths carry unknown provenance, and after a relaunch the ledger is empty (`entries:0` above) so `isMirroredCopy` cannot tell the user's file from a materialised copy that outlived the disconnect meant to delete it. Pushing on that guess would **overwrite a remote that may have moved on**.

So seeded paths are gated on the remote — the option you picked:

- **absent on the remote** → nothing we materialised could be its source, so it is the user's → **pushed**
- **present on the remote** → we cannot tell which side is newer → **left alone**, and *logged*, because a file quietly not pushed is the exact shape of bug this area keeps producing

An edit from then on is an ordinary event and takes the ordinary path, ledger and all.

`remoteExists` is **optional** and defaults to "assume present", so a deployment that does not wire it keeps today's behaviour rather than gaining a push nobody asked for.

## Known limitation, stated rather than hidden

A file that exists on both sides and was genuinely edited out-of-band *while disconnected* is not pushed at connect. Distinguishing that from a stale mirror needs a persisted materialisation ledger (the ledger survives a clean disconnect but not a crash). Worth its own issue; not smuggled in here.

## Verification

- `tsc --noEmit` (src) — clean; `e2e/**` — clean
- vitest — **1317 passed**, 86/86 files, including 2 new `LocalWriteWatcher` cases pinning both halves of the gate

Two pre-existing watcher tests initially went red against a first draft of the harness default; the default was wrong, not them — their starting `files` mean "the user's content, to be pushed", and the harness now says so. **No existing assertion was changed.**

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.25",
+  "version": "1.1.8-beta.26",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/manifest.json
+++ b/plugin/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.25",
+  "version": "1.1.8-beta.26",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/package-lock.json
+++ b/plugin/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "obsidian-remote-ssh",
-  "version": "1.1.8-beta.25",
+  "version": "1.1.8-beta.26",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "obsidian-remote-ssh",
-      "version": "1.1.8-beta.25",
+      "version": "1.1.8-beta.26",
       "license": "MIT",
       "dependencies": {
         "@xterm/addon-fit": "^0.11.0",
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "obsidian-remote-ssh",
-  "version": "1.1.8-beta.25",
+  "version": "1.1.8-beta.26",
   "description": "VS Code Remote SSH-like experience for Obsidian",
   "main": "main.js",
   "scripts": {
--- a/plugin/src/adapter/LocalWriteWatcher.ts
+++ b/plugin/src/adapter/LocalWriteWatcher.ts
@@ -38,6 +38,16 @@ export interface LocalWriteWatcherDeps {
    * never — a deployment without the disk cache is unaffected.
    */
   isMirroredCopy?(rel: string, stat: LocalFileStat): boolean;
+  /**
+   * Whether the remote already has this path. Consulted ONLY for files
+   * found by the startup scan, whose provenance is unknown: a note a
+   * plugin wrote while nothing was watching, or a materialised copy that
+   * outlived the disconnect that should have removed it. Absent from the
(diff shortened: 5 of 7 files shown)


sotashimozono:
## Result: the bug this PR is about is fixed

`fs-visibility:773 REAL-PLUGIN REPRODUCTION` and `plugin-code-roundtrip:344 push:` are both gone from the failure list on [run 30784185008](https://github.com/sotashimozono/obsidian-remote-ssh/actions/runs/30784185008).

| shard | #507 | #506 |
|---|---|---|
| features | ✅ 27 passed | ✅ 27 passed |
| stress | **19 passed** / 2 failed / 1 flaky | 18 / 3 / 1 |
| core | **18 passed** / 2 failed / 1 flaky | 19 / 1 / 2 |

`:773` failed on both #505 and #506 and passes here, so the fix is doing what it claims: a plugin's `fs.writeFileSync` in `onload()` now reaches the remote.

## Correction to the description

I claimed `plugin-code-roundtrip:344 push:` was a "real regression — the storm was carrying it". **That is wrong.** Checking every run rather than the two I had looked at:

| run | `:344` |
|---|---|
| #504 (no #490) | ✓ 3.8 s |
| #505 (with #490) | ✘ |
| #506 (no product change) | **✓ 4.6 s — and ✓ on retry** |
| #507 | ✓ |

It passed on #506 with no product change between #505 and #506, so it is **flaky**, not a deterministic #490 regression. The table in the description overstates it. `:773` is the load-bearing evidence for this PR and it is unaffected by the correction.

## The new core red is NOT the downgrade scenario

`plugin-code-roundtrip:419 an OLDER local copy does not DOWNGRADE a NEWER remote plugin` reads alarmingly given what this PR changes, so: it is not the gate misfiring. The failure is in setup, before the downgrade assertion runs:

```
Error: shadow vault did not load the remote tree
  (need getMarkdownFiles()>=1 AND a file-explorer leaf) within 30000ms.
shadow vault model: {"fileMapKeys":1,"rootChildren":0,"loadedFiles":1,"markdownFiles":0,"fileExplorerLeaves":1}
```

and the plugin log for that run shows connect succeeding normally —
`populateVaultFromRemote(shadow-layout-ready): rpc-walk, 5 entries (2 hidden) in 41ms [lazy: root level]` — with the model ending up at `markdownFiles: 0` anyway. It was already flaky on #506 (failed once, passed on retry) before this change existed. Separate issue; plugin code lives under `.obsidian/plugins/`, which `makeLocalWriteIgnore` excludes from the watcher entirely.

sotashimozono:

PR #508, opened 2026-08-03
test(e2e): print the cold-read evidence that Playwright was discarding

sotashimozono:
Stacked on #507.

## My own instrumentation threw the evidence away

#506 attached the full materialisation dump to the two cold-read reds. [Run 30782633453](https://github.com/sotashimozono/obsidian-remote-ssh/actions/runs/30782633453) printed **none of it**. The reason is in the assertion:

```ts
expect(read.error, '<dump>').toContain('ENOENT');
```

When the read **succeeds** — which is the failure being investigated — `read.error` is `null`, and `toContain` on null raises Playwright's own:

```
Matcher error: received value must not be null nor undefined
Received has value: null
```

That error **discards the custom message**. So the one run that could have named the cause threw the evidence away, and the diagnostic I added to settle this looked like it had nothing to say. That is on me.

## The fix

Both sites now pass `read.error ?? '<no error: the read SUCCEEDED, returning …>'`.

Identical strength — a successful read still fails, because that description does not contain `ENOENT` — and the message survives. It also shows *what* came back, which distinguishes "the real note" from "something else at that path".

## The dump gained what it was missing for this question

- **mtime per file.** "What is on disk" does not say *when* it landed, and the whole question is whether a file appeared before or after the thing that supposedly asked for it. Rendered ISO, to line up against the log's timestamps.
- **an unfiltered 10-line log tail.** The filtered view answers *"did the watcher act"*. It cannot answer *"then what DID put this file here"* — and the writer may be `populateVaultFromRemote`, `preSpawnPull`, `installMissingShadowPlugins`, or something unlooked-for.

**No assertion, matcher or timeout is changed.**

## Where the stack stands

| shard | #507 |
|---|---|
| features | ✅ 27 passed |
| stress | 19 passed / **2 failed** (`:494`, `:578` — this PR's target) / 1 flaky |
| core | 18 passed / 2 failed (`fs-sync-bridge-probe`; `:419` shadow-vault-load flake) / 1 flaky |

#507 fixed the real one: `fs-visibility:773 REAL-PLUGIN REPRODUCTION` — a plugin's `fs.writeFileSync` in `onload()` now reaches the remote.

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.26",
+  "version": "1.1.8-beta.27",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/fs-visibility.spec.ts
+++ b/plugin/e2e/fs-visibility.spec.ts
@@ -520,8 +520,15 @@ test.describe('a plugin\'s view of the filesystem must be the REMOTE vault (#429
     // Cold: nobody has asked for this note, so the synchronous surface
     // has nothing to hand over. This is the CONTRACT, not a defect —
     // the async test below is the half that fetches.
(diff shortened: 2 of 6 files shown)


sotashimozono:

PR #509, opened 2026-08-03
test(e2e): probe whether the metadata index can come from the remote

sotashimozono:
Stacked on #508. First step of the **B1 + B3** direction: serve the index from the server, keep only the index locally.

## Why a probe and not an implementation

The requirement is that note **content** lives on the server and the wire is not paid for it. Splitting the vault into three layers:

| layer | what | today |
|---|---|---|
| **A. tree** | names, paths, sizes, mtimes | ✅ remote walk, **zero bytes of content** (`BackgroundIndexer`) |
| **B. index** | links, tags, headings, frontmatter | ❌ Obsidian rebuilds it here, **by reading every note** |
| **C. content** | the bytes | ✅ on demand only, bounded cache, deleted on disconnect |

B is the whole problem, and it is not our code doing it: the moment a file lands in `vault.fileMap`, Obsidian reads its content through our patched adapter to parse it. `BackgroundIndexer.ts:78` records where this was left:

> File CONTENT is never read — only the tree. (Obsidian's own metadataCache reads the files it cares about; **that traffic is its call, not ours**.)

Whether we can take that traffic back is **not answerable from the typings**. `obsidian.d.ts` declares `getFileCache` / `getCache` and no setter — but says nothing about what the running app exposes. So this probes it, exactly as `fs-sync-bridge-probe.spec.ts` already does for the fs question.

## What it measures, and what each answer decides

1. **The cost, in bytes.** `readCache.stats()` after connect — and nothing in the spec opens a note, so that is precisely what Obsidian pulled for indexing. Plus how many markdown files ended up with parsed metadata. *This is the number the whole decision turns on, measured instead of assumed.*
2. **Is there anywhere to put an index?** `metadataCache` own keys, prototype methods, and which of them are path-keyed objects.
3. **Do `resolvedLinks` / `unresolvedLinks` drive the link graph?** They are the one pair the public typings expose as plain **writable** records. If backlinks follow what is written there, links/backlinks/graph can be served remotely **even if `CachedMetadata` cannot be injected** — that would make B1 partially reachable regardless.
4. **Does Obsidian persist metadata?** A store keyed by `(path, mtime, size)` is the shape where a pre-seeded entry lets Obsidian skip the read entirely — the cleanest possible B1+B3.

**Key names only, never values.** A vault's metadata is the user's content and this runs in CI.

## What it asserts

Only that it measured something: `app` was reachable, and the vault was not empty. A probe that quietly returns nothing is worse than no probe — it reads as an answer.

Runs in `features` (2 min, green) rather than `core`, which already carries the other probe and is the slowest shard. Verified with `--list`: 27 → 28 tests, 3 → 4 files.

## Stack status (#508)

| shard | result |
|---|---|
| features | ✅ 27 passed |
| stress | 19 passed / 2 failed (`:494`, `:585`) / 1 flaky |
| core | 19 passed / 1 failed (`fs-sync-bridge-probe`) / 2 flaky |

The two stress reds are **this design question**, not a defect: they assert that a note nobody asked for is not synchronously readable, and Obsidian's indexer has already asked for everything. They stay red until B1/B3 lands — going green any sooner would mean weakening them.

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.28",
+  "version": "1.1.8-beta.29",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.27",
+  "version": "1.1.8-beta.28",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.29",
+  "version": "1.1.8-beta.30",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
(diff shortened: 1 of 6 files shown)


sotashimozono:

PR #510, opened 2026-08-03
fix(e2e): stop the fs probe deadlocking the renderer it is probing

sotashimozono:
Stacked on #509.

`fs-sync-bridge-probe` has been `core`'s standing red for this whole stack, and the 300 s hook budget #504 gave it changed nothing — **the hook was never the problem.**

## Cause 1: the probe deadlocks the renderer, on purpose, every run

Section 7 sent a **synchronous XHR to the ResourceBridge**:

```ts
xhr.open('GET', bridgeUrl, false);   // false = synchronous
xhr.send();
```

The bridge's HTTP server runs on the renderer's **own event loop**. Blocking there blocks the thread that would accept the connection — the reply is not slow, it is *unreachable*. `xhr.timeout` cannot bound it either: setting it on a synchronous XHR throws. So the probe wedged the renderer, the test hit its budget, and CI reported a bare `Test timeout of 120000ms exceeded` with no report attached.

**The spec already knew this.** The comment under its assertions describes the self-deadlock exactly — and keeps the request anyway, "because the numbers are the evidence any future 'just do it synchronously' idea has to answer for". Those numbers were never going to arrive.

- `syncXhrAgainstBridge` now records the **structural reason** and sends nothing. That is the stronger argument, and it is what the product already relies on (`materializeSync` serves only bytes already in the read cache).
- `syncXhrPermitted` keeps the part that *is* measurable — whether this Chromium still permits a main-thread synchronous XHR at all — against a `data:` URL that does not need this event loop to answer.

## Cause 2: the budget was on the wrong thing

`test.setTimeout(300_000)` in `beforeAll` bounds the **hook**, not the body. Sections 5 and 6 spawn five subprocesses and make five remote reads, which is not a comfortable 120 s on a cold container. The body now gets its own 240 s.

## No assertion changes

`syncXhr` was already deliberately not asserted. The four preconditions — `hasRequire`, `fsModuleCacheShared`, `fsPatchTakesEffect`, `hasChildProcess` — are untouched.

## Verification

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.30",
+  "version": "1.1.8-beta.31",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/fs-sync-bridge-probe.spec.ts
+++ b/plugin/e2e/fs-sync-bridge-probe.spec.ts
@@ -80,6 +80,11 @@ test.afterAll(async () => {
 
 test.describe('capability probe: can Node fs be routed to the remote daemon? (#429)', () => {
   test('PROBE — report the renderer capabilities that decide the design', async () => {
+    // The 300 s in `beforeAll` bounds the HOOK; the body still had the default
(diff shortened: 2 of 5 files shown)


sotashimozono:
Closing: the spec this fixes, `fs-sync-bridge-probe.spec.ts`, was deleted in #541 along with the raw-fs direction it was probing. Its own header said it was "not a product test".

PR #511, opened 2026-08-03
test(e2e): find out whose fault the metadata re-read is

sotashimozono:
Stacked on #510.

## Where the probe got to

With `indexingSettled: true` on **both** runs — so no more "we killed it before it saved":

```json
"firstRunNoteBytes": 929, "secondRunNoteBytes": 929, "secondOverFirst": 1,
"cachedRecordsOnSecondRun": 11
```

`totalMdBytes` is also 929, so the ratio is exactly **1.000** once `.obsidian` traffic is separated out: **Obsidian reads 100% of note content on every start.** And it does so despite having 11 records persisted in this vault's IndexedDB.

The verdict called that *"reuse is refused, so the index must be injected at runtime"*.

## That skips a possibility — and it is the cheap one

Obsidian validates a cached entry against the file's **identity**. If **our** side reports a different `(mtime, size)` on each launch, every entry looks stale and the re-read is **our own doing**. That is fixable, and seeding the cache would then work — a much smaller build than runtime injection into `fileCache` / `metadataCache` / `resolvedLinks`.

So the probe now fingerprints `TFile.stat` per markdown file and compares the two launches:

- `statStableAcrossLaunches` — did the identities hold
- `statChanged` — what moved, capped at ten entries
- the verdict grows a **fourth** outcome, one that names us rather than Obsidian

**This is not a claim that this is the cause.** It is the check that has to happen before "runtime injection" is treated as settled, because the two answers cost very different amounts to build.

## Verification

Refs #429

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.31",
+  "version": "1.1.8-beta.32",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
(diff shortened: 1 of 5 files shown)


sotashimozono:

Issue #513, opened 2026-08-03
Serve the metadata index from the remote (#429 follow-up): Obsidian re-reads 100% of note content on every start


sotashimozono:
Follow-up to #429. **Measured, not assumed** — every number below comes from `e2e/metadata-index-probe.spec.ts` on [run 30802603656](https://github.com/sotashimozono/obsidian-remote-ssh/actions/runs/30802603656).

## The requirement

Note **content** lives on the server. The local shadow vault holds a minimal set of files. The reason not holding the whole vault is a problem at all is that hyperlinks, Dataview and graph misbehave — so the *index* has to exist, even though the *content* must not.

## Three layers

```mermaid
flowchart LR
  subgraph R["remote (server)"]
    C["C: content<br/>GB scale"]
  end
  subgraph L["local (shadow vault)"]
    A["A: tree<br/>names / paths / sizes / mtimes"]
    B["B: index<br/>links / tags / headings / frontmatter"]
    D["D: materialisation cache<br/>128MB / 8MB, deleted on disconnect"]
  end
  C -->|"walk — zero content bytes"| A
  C -->|"❌ Obsidian reads EVERY note"| B
  C -->|"on request only"| D
  B --> F["hyperlinks / Dataview / graph"]
```

| layer | today |
|---|---|
| **A. tree** | ✅ `BackgroundIndexer` walks names, reads no content |
| **B. index** | ❌ **Obsidian rebuilds it here, by reading every note** |
| **C. content** | ✅ on demand only, bounded, deleted on disconnect |

B is not our code doing it: the moment a file lands in `vault.fileMap`, Obsidian reads its content through our patched adapter to parse it. `BackgroundIndexer.ts:78` already recorded the position this was left in — *"Obsidian's own metadataCache reads the files it cares about; that traffic is its call, not ours."*

## The knot

Everything in `vault.fileMap` is what makes links, Dataview and graph work — **and it is exactly what makes Obsidian read every note**. Fewer files in the model means less transfer and broken links; all of them means links work and the whole vault crosses the wire.

### The scaling law

```
markdownInModel: 5,  withParsedMetadata: 5      → Obsidian parses 100% of the markdown
noteBytes: 929 / totalMdBytes: 929              → ratio exactly 1.000
```

[...]

PR #514, opened 2026-08-03
test(e2e): can we keep the tree and refuse the content?

sotashimozono:
Stacked on #512. First step of the lazy-filesystem direction — and it may cut how much has to be given up.

## Injecting the index is off the table

Measured on [run 30802603656](https://github.com/sotashimozono/obsidian-remote-ssh/actions/runs/30802603656), with all four alternatives closed:

| check | result |
|---|---|
| killed before it saved? | `indexingSettled: {firstRun: true, secondRun: true}` |
| does it persist? | **yes — 11 records in this vault's IndexedDB** |
| do we hand it a different `(mtime, size)`? | **no — `statStableAcrossLaunches: true`, `statChanged: []`** |
| re-reads anyway? | **yes — 929/929, ratio 1.000** |

> `persisted, identity stable, and re-read anyway — reuse is genuinely refused`

Seeding cannot help. The only route left was writing into undocumented internals. Full write-up in #513.

## But "give up on the index" concedes more than necessary

The tree and the content are not equally expensive and do not buy the same things. `BackgroundIndexer.ts:55` records the free half — `metadataCache` resolves links against **`fileMap` alone**:

| feature | needs |
|---|---|
| file tree, open, edit, write | **tree only** |
| `[[link]]` resolve / click / suggest | **tree only** |
| Quick Switcher | **tree only** |
| graph **nodes** | **tree only** |
| tags / frontmatter / link extraction | content |
| backlinks | content |
| global search | content |
| Dataview | content |

Today they arrive as a package, because *registering* a file is what makes Obsidian read it.

## The one documented lever that might separate them

Obsidian's own **"Excluded files"** setting — `userIgnoreFilters` on disk, `isUserIgnored` / `updateUserIgnoreFilters` on the running `metadataCache`, all three of which the shape probe already found in `ownKeys`/`protoMethods`.

Unlike writing to `fileCache` / `metadataCache`, **this is published, user-facing behaviour** — a design resting on it is not a bet on internals.

So: set the filter, relaunch, measure. **Two outcomes, reported separately:**

- did the content reads stop?
- did the files **survive in the model**? An exclusion that empties `fileMap` takes links and Quick Switcher with it and is worth nothing.

It deliberately does **not** use `waitForShadowVaultLoaded` — that requires `getMarkdownFiles() >= 1`, and "the exclusion emptied the model" is one of the answers being measured, not a reason to abort before reporting it.

## If it comes back positive

Links, Quick Switcher and graph nodes stay at **zero content transfer**, at GB scale, and only backlinks / search / Dataview need an α (daemon-side grep and reverse-index, shown in our own views). If it comes back negative, lazy registration is the only dial left and the scope question is how deep to go.

Either way `vault-features.spec.ts` 2, 6, 7, 9 are affected — that is a **spec change**, not a test to weaken, and it is yours to sanction.

Refs #429, #513

### DIFF
--- a/manifest-beta.json
+++ b/manifest-beta.json
@@ -1,7 +1,7 @@
 {
   "id": "remote-ssh",
   "name": "Remote SSH",
-  "version": "1.1.8-beta.33",
+  "version": "1.1.8-beta.34",
   "minAppVersion": "1.5.0",
   "description": "Edit remote vaults over SSH/SFTP — VS Code Remote-SSH style.",
   "author": "souta shimozono",
--- a/plugin/e2e/metadata-index-probe.spec.ts
+++ b/plugin/e2e/metadata-index-probe.spec.ts
@@ -8,6 +8,8 @@ import {
 import { scaffoldTestVault, type ScaffoldResult } from './helpers/vault-scaffold';
 import { assertSshdReachable } from './helpers/sshd';
 import { logPathFor } from './helpers/log-oracle';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
 
 /**
  * CAPABILITY PROBE — can the link/tag index be served from the remote, instead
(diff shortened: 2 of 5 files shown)


sotashimozono:

PR #535, opened 2026-09-14
feat(vault): allow selected dot-folders per SSH profile

### LINKED ISSUE #534: Feature request: allow selected dot-folders per remote profile
### Problem / motivation

I use Remote SSH to read and edit Markdown documentation in Git worktrees on a remote server. Some worktrees live beneath a dot-prefixed directory, for example:

```text
/home/user/.herdr/worktrees/project/my-branch/docs/writing/article.md
```

With `/home/user` as the remote vault root, `.herdr` and its documents do not appear in the file explorer. The same file is accessible through VS Code Remote SSH. This occurred with Remote SSH **1.1.7**, Obsidian **1.13.7**, on macOS using **SFTP**.

Opening the `docs/writing` directory directly as a separate remote vault works, but requires another profile and loses navigation across the surrounding worktrees.

I understand that #450 deliberately introduced dot-folder filtering to reduce clutter and avoid walking large caches. The current `BulkWalker` also filters entries whenever any vault-relative path segment begins with a dot. That default is useful, but it also excludes intentionally hidden directories containing documents users want to edit.

### Proposed solution

Add a **per-profile opt-in allowlist for selected dot-folders**, ideally using vault-relative paths. For example, allowing `.herdr` would make its ordinary descendants available for navigation and Markdown editing from the home-directory vault.

Suggested behavior:

- Keep the current hidden-by-default behavior for profiles without an allowlist.
- Allow an explicitly selected dot-folder and its ordinary descendants through the visibility filter, consistently for initial loading, lazy folder expansion, and background indexing.
- Continue honoring the existing Ignore directories setting. Allowing `.herdr` should not implicitly expose or traverse nested `.git`, dependency caches, or every other dot-folder.
- Keep the vault configuration directory and per-device state excluded from the note tree.
- Explain the filtering in the settings UI so users can understand why an existing remote directory is absent and how to include it.

Acceptance example: with a remote root of `/home/user` and `.herdr` allowed, I can navigate to `.herdr/worktrees/project/my-branch/docs/writing/article.md` and edit the existing remote file without creating another profile. With no allowlist configured, behavior remains unchanged.

A selected-folder allowlist would meet this use case; a global “show all hidden files” switch is not necessary.

### Expected tier

Either

JacobStephens2:
## Summary

- A vault rooted at `/home/user` currently hides Markdown documents beneath `.herdr/worktrees`. Add **Allowed hidden directories** to each SSH profile so users can opt in to exact vault-relative dot-folder paths and navigate to those documents without a symlink or a separate vault. Implements #534.
- Apply the same visibility rules to full walks and lazy folder expansion, which also supplies the background indexer. Hidden names remain excluded by default; each nested dot-folder needs its own entry, dotfiles remain hidden, and ignored directories plus the vault configuration remain excluded. Prune excluded SFTP directories before descending so allowing a worktree directory does not trigger walks through nested caches.
- Document the setting and replace the empty-vault notice's rename suggestion with guidance to the profile settings. Bump the beta version to **1.1.8-beta.13**, leaving stable manifests unchanged.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] CI / tooling
- [ ] Dependency update

## Test plan

- [x] Existing tests pass (`npm test`): **1,319 passed**, 2 existing TODOs, across 86 files.
- [x] New tests cover successive SFTP expansion through `.herdr`, exact nested allowlist paths in RPC walk results, configuration exclusions, and ignored-directory precedence with no SFTP descent into excluded trees.
- [x] Manually verified the **1.1.7 backport of this change** in Obsidian **1.13.7** on macOS against a live SSH server using **SFTP**: opened a home-directory vault, expanded `.herdr/worktrees/.../docs/writing`, found the article in the quick switcher, and opened the existing remote Markdown document in Live Preview. No symlink was used for that path.

- [x] Manually verified the **1.1.7 backport** using **RPC** with the official 1.1.7 daemon: a server-created note appeared under the allowed `.herdr` directory, a server-side edit updated the open note, and deleting the disposable note removed it from the file tree without reconnecting or re-expanding the folder.
- [x] Added `populateVaultFromRemote.reconnect.test.ts`: seed an already-expanded folder in the retained vault model, add a remote file before reconnect (without emitting a watch event), and invoke the actual reconnect populate method. With background indexing disabled, the root-only control misses the file. The branch's normal RPC and SFTP paths automatically index it, attach it to the folder, and emit one create event. The tests exercise the real populate, walker, visibility filter, indexer, and model builder with simulated remote I/O and Obsidian UI.

The installed 1.1.7 backport required re-expanding a folder to discover a file already present before daemon startup. That specific missing-file scenario did not reproduce with this branch's background indexer in the controlled tests; no reconnect implementation change was added. This is not a live beta installation test. The full Docker integration/E2E suites were not run locally.

## Checklist

- [x] `tsc --noEmit` passes.
- [x] `npm run lint` completes with 0 errors (3 warnings: the existing settings-search warning and two hardcoded-default-config-name warnings; the active custom `configDir` is also passed to the filter and tested).
- [x] Bundle size within limit (`npm run build`): **900,623 bytes**, below 921,600 bytes.
- [x] Beta manifests agree, the version exceeds `next`, and stable manifests/compatibility maps are unchanged.
- [x] No secrets or credentials committed; the public diff contains source, tests, documentation, and beta version metadata.

Closes #534.


sotashimozono:
Thanks a lot for this, and sorry it sat for a week.

The design is right: dot-names stay hidden by default, allowances are exact-path and directory-only, and `.obsidian` / ignored dirs can never be re-exposed. Pruning excluded dirs before descending on the SFTP path is a nice bonus.

All checks pass. The E2E jobs are red on `next` independently of this PR (known #429-related specs), so I'm merging without waiting for them.

Merging — this ships as 1.1.8-beta.13.
Please wait for 1.1.8 release.

PR #546, opened 2026-09-22
test(e2e): time every vault.readBinary in the scale benchmark (beta.19)

sotashimozono:
## Why

The first scale runs (#545) show a flat **~43 ms per note** (about 23 notes/s) for the metadataCache index: 1,000 notes in ~54 s, 10,000 in ~460 s. It is the same on LAN and on a 40 ms WAN, and the note bytes arrive up front. From Obsidian 1.13's `app.js`, `metadataCache` indexes through a **sequential** `workQueue`, one note at a time: `vault.readBinary`, then hash, then `postMessage` to the worker. Our adapter's cold `readBinary` is two RPCs (`readBinary` + `stat`).

Before optimising either side, measure which part the 43 ms is.

## What

The scale spec wraps `app.vault.readBinary` on the first sample, before the remote tree lands. For each launch it records the read count, the average and max latency, and the peak number of reads in flight. `1` means strictly sequential. These go in the per-sample log line, the JSON artifact and three new summary-table columns.

Experiment. Draft until the numbers are in.

## Test plan

- [x] `tsc --noEmit`
- [ ] `Scale (s / lan)`, `Scale (s / wan)` summaries show the read columns


sotashimozono:
## The harness was measuring its own bug

`registerVault` minted a fresh `crypto.randomBytes(8)` id for the vault on **every launch**. Obsidian uses that id as `app.appId`, and its metadata cache lives in IndexedDB under `<appId>-cache`:

```
cold     appId a8964d01349aea18   databases: 4
relaunch appId 719c40771cd85e02   databases: 6   ← a new pair every launch
```

So every launch started from an empty cache, and the databases piled up. That is why a relaunch re-read every note, and **#513's "persisted, identity stable, and re-read anyway" was measured on this same harness**. Fixed here: the id is derived from the vault path, and an existing entry for that path is reused.

## What the numbers say once the id is stable

`s` (1,000 notes, ~5 MB), LAN, relaunch pass:

| | cache at start | reads | tx | parsed |
|---|---|---|---|---|
| this branch (no tree restored) | 1003 entries, 1001 hashed | 1001 | 7.5 MB | 94 s |
| with #547 (tree restored in onload) | 1003 entries, 1001 hashed | **0** | **0.1 MB** | **13 s** |

The cache survives either way. Without the tree in the model, `metadataCache.initialize()` deletes it and Obsidian reads everything again — which is exactly what #547 fixes.

Also in this branch: reads are strictly sequential (`inflightMax=1`) at ~42 ms each, shutdown is now a real window close rather than SIGTERM, and each sample reports `fileCache` size and hashed entries.

PR #547, opened 2026-09-22
feat(vault): restore the remote tree at startup so Obsidian keeps its index (#513) (beta.20)

sotashimozono:
## Why

The scale benchmark (#545) showed every note being read and parsed again **on every launch**: about 23 notes/s, so 10,000 notes (~50 MB) take ~460 s, on a relaunch as much as on a cold start. Reading Obsidian 1.13's `app.js` found the cause, and it is ours to fix.

```
plugins.initialize()          ← our onload (awaited by Obsidian)
vault.load()                  ← just starts adapter.watch
metadataCache.initialize()    ← for each cached path NOT in the vault model: deletePath()
                                 → also deleted from IndexedDB
                                 for each path whose stat matches the cache: kept, not read
layout ready → connect → we register the remote tree → nothing cached → every note read, one by one
```

The shadow vault's disk has no notes, so `initialize()` threw the whole index away every time. This is #513's "persisted, stat stable, re-read anyway".

## What

- **`TreeSnapshot`** (`vault/TreeSnapshot.ts`): after a background pass over the daemon's `fs.walk` (real `mtime`/`size`), save the tree to `~/.obsidian-remote/state/<profile>/tree-snapshot.json`. That's paths, mtime and size only, no content, about 50 bytes per note, written atomically. It sits next to `community-plugins.base.json`: per-device, never synced, outside every vault.
- **Restore in `onload`**: when this is a shadow vault (`autoConnectProfileId`) in lazy mode and a snapshot exists for this profile and `remotePath`, build it into the model before `onload` resolves. `metadataCache.initialize()` then finds every note with the stat it cached and **reads none of them**. The File Explorer also shows the tree before connect.
- **`BackgroundIndexer` reconciles** (`modelAtStart` / `currentStat` / `onComplete`), as well as filling in:
  - A complete daemon walk `modifyOne`s notes whose mtime/size changed and `removeOne`s paths the remote no longer has.
  - It only touches entries that were present when the pass started, so a note created mid-walk is never removed.
  - It skips a stat update if a live `fs.watch` event already changed the note.
  - It removes nothing after a truncated or cancelled walk.
  - SFTP fallback: removals only, and only if every folder listed; no stat comparisons, because its stats are 0.
  - If the model can't be read, it falls back to filling in only.
- **Post-connect catch-up**: a restored note with no metadata had its read fail on the empty shadow disk before our adapter was patched. Those are `modify`d once, so they go through the read queue over SSH.
- **SFTP safety**: a snapshot restored before an SFTP-fallback session carries a daemon session's stats, and metadataCache would trust them as "unchanged". After such a pass, those notes' stats are zeroed (forcing a re-read, which is today's behaviour) and the snapshot is dropped.
- `sanitiseStateKey` exported for the state path.

`1.1.8-beta.20`

## Risks / open questions

- **Before connect, the tree exists but content doesn't.** Obsidian's native adapter serves reads from the empty shadow disk until the patch. A restored tab or an early click fails to read until connect. I haven't seen Obsidian delete a model entry after a failed read yet; the E2E restart specs on this PR will tell.
- A stale snapshot costs a re-read of what changed, never wrong state: reconcile fixes the model, and a changed stat makes metadataCache re-read.
- Non-lazy (`lazyFolderLoad: false`) is untouched. No snapshot is restored, because nothing would reconcile it.

## Test plan

- [x] New `TreeSnapshot.test.ts` (round trip, wrong remotePath, corrupt/version, atomic write, model collection skips root and configDir)
- [x] New `BackgroundIndexer.reconcile.test.ts` (12 cases: stat update, removals incl. folder cascade, mid-walk create kept, live-changed stat skipped, truncated, cancelled, opt-out, unreadable model, SFTP removals-only, SFTP failed folder)
- [x] `vitest run`: 1283 passed · `tsc` · lint 0 errors · build
- [ ] E2E core / features / stress (restart-settings, sync, reflect, vault-features)
- [ ] Scale benchmark on this branch (run 35754461290, `s,m` × `lan`): the relaunch pass should show near-zero reads and parse time


sotashimozono:
## Measured: it works

`s` (1,000 notes, ~5 MB) over LAN, **relaunch** pass, with the harness's vault-id bug fixed in #546:

| | reads | bytes | every note parsed |
|---|---|---|---|
| without this PR | 1001 | 7.5 MB | 94 s |
| **with this PR** | **0** | **0.1 MB** | **13 s** |

At `t=3s` on the relaunch the model already holds all 1,001 notes and 300 are parsed; by `t=13s` all of them are, with `reads=0`. The cold pass is unchanged (~54 s), as expected: it still reads every note once, and now it also writes the snapshot (`TreeSnapshot: saved 1111 entries`).

Both runs start with the same cache (1003 entries, 1001 hashed). The difference is only whether the tree is in the model when `metadataCache.initialize()` runs: without it, Obsidian deletes the lot and re-reads; with it, it keeps the lot and reads nothing.

`m` (10,000 notes) is running now.

PR #548, opened 2026-09-22
perf(rpc): send each frame in one write on both ends (beta.21)

sotashimozono:
Stacked on #546, which adds the read-timing columns this PR is measured with.

## Why

#546 measured metadataCache's per-note read. It is **strictly sequential** (`inflightMax=1`), and each `vault.readBinary` takes **~41 ms on LAN and ~42 ms on a 40 ms WAN**: a fixed cost, not the link. At 10,000 notes that's ~460 s per launch.

Both ends write a frame as two writes, header then body:
- plugin `FramedDuplex.writeMessage` (`transport/framing.ts`)
- daemon `rpc.WriteFrame` (`server/internal/rpc/frame.go`)

The daemon listens on a unix socket that sshd forwards over the SSH connection. sshd sends each chunk it reads as its own SSH packet, and for a non-tty session its TCP socket has Nagle on (no `TCP_NODELAY`). So the body packet waits for the client's delayed ACK of the header packet, which Linux holds for up to **40 ms**. That happens on every reply, and symmetrically on requests.

## What

Build `header + body` into one buffer and write it once, on both ends. Single-write framing is pinned in a test on each side (`framing.test.ts`, `TestWriteFrame_SingleWrite`). Wire format unchanged, so old and new daemons and plugins interoperate.

`1.1.8-beta.21`

## Test plan

- [x] `vitest run tests/framing.test.ts`: 11 passed
- [ ] `Server build & test` (Go tests, including the new single-write test; I had no local Go)
- [ ] Scale `s` × `lan,wan` on this branch: `avg read` should drop from ~41 ms to a few ms, and the parse time with it


sotashimozono:
## It works — the earlier "no effect" reading was on the wrong measurement

The Node-side RPC latency test (added in #546) gives a clean A/B against the same daemon and SSH transport, 30 sequential calls each:

| branch | `fs.stat` | `fs.readBinary` |
|---|---|---|
| #546 (two writes per frame) | avg **41 ms**, min 41, max 42 | avg **41 ms**, min 41, max 42 |
| **this PR** (one write) | avg **0 ms**, min 0, max 1 | avg **0 ms**, min 0, max 1 |

A 41 ms floor with no spread is not a network: it is Nagle waiting on the peer's delayed ACK, which Linux holds for 40 ms. One write per frame removes it, and the round trip becomes sub-millisecond.

Earlier I said this PR "did not move it" based on the scale benchmark, which measures through Obsidian's renderer and still showed ~42 ms per `vault.readBinary`. That reading stands for the renderer and is not yet explained — the E2E does stage the locally built daemon (`server-bin/.dev-daemon`, shared into the shadow vault), so both ends should have the fix there too. A scale run on this branch with the adapter probe from #546 is running now; its `adapter probe: {"statAvgMs": …}` will say whether the renderer path is still at 41 ms.

Either way the round trip itself is fixed, and every RPC this plugin makes pays it.

sotashimozono:
## Follow-up on the review's memory finding: not acting, and why

The release review flagged that `Buffer.concat([header, body])` copies the whole body per frame, doubling transient memory for in-flight large frames, with Obsidian Mobile as the concrete failure mode.

Checked, and the suggested remedies don't hold here:

- **Mobile is not a case.** `plugin/manifest.json` has `"isDesktopOnly": true` — this plugin never loads on mobile. What remains is a 16 MB frame transiently costing 32 MB on a desktop, which is not a failure mode.
- **`cork()`/`uncork()` would undo the fix.** Node coalesces corked writes only when the stream implements `_writev`; without it, `Writable` falls back to one `_write` per chunk. `ssh2`'s `Channel` (1.17.0) does not implement `_writev` — so corking would go back to two writes per frame, and with it the 40 ms Nagle + delayed-ACK stall this PR removed (41 ms → under 1 ms per round trip).
- **`net.Buffers` on the daemon** would use `writev` on the unix socket, but fixing one direction while the client still concatenates buys nothing measurable, and the same 16 MB ceiling applies.

So the copy stays. Recorded here rather than silently dropped.

Issue #519, opened 2026-08-10
Wikilink from vault-root file to sub-repo file does not navigate (fileMap appears lazy)


Justin318:
## Summary

Wikilinks authored in a vault-root markdown file (e.g. a dashboard / hub note) that point to files inside nested subfolders do not navigate when clicked. Expected: opening the link fetches the target via the adapter (like opening any other file). Actual: click is a no-op (or renders a "Create new note" prompt), even though the target file exists on the server.

This breaks a common dashboard / hub pattern where the root note links out to project notes, plan files, and other artifacts scattered across independent sub-repository folders within the same vault.

## Environment

- Plugin: `obsidian-remote-ssh` (latest main as of 2026-08-10)
- Transport: RPC (recommended)
- Obsidian: desktop, vault mounted via this plugin
- Server layout: vault root contains both regular files (e.g. `README.md`, a dashboard note) and nested subfolders that are themselves independent git repositories (each with its own `.git/`). This is a common multi-repo workspace layout where a single tool-of-tools workspace holds several projects side-by-side.

## Repro

1. Open Obsidian on a freshly created profile pointing at the remote vault. Do **not** manually click into any subfolder in the File Explorer.
2. Open the vault-root dashboard file.
3. Click any wikilink that targets a file inside a nested subfolder (e.g. `[[<subfolder>/<project>/README.md|...]]` or `[[<subfolder>/plans/<plan>.md|...]]`).

Expected: target file is fetched via the adapter and rendered in the editor.

Actual: nothing happens, or Obsidian shows the "Create new note" prompt because the target is not in `app.vault.fileMap`. The same wikilink works after manually expanding the target folder in the File Explorer once per session — i.e. after the underlying `TFile` has been lazily registered.

## Hypothesis

`VaultModelBuilder` walks the remote tree at startup but only registers files into `app.vault.fileMap` for folders Obsidian has actually visited in the File Explorer. Wikilink resolution then consults `fileMap` first; an unvisited subfolder's files are not in the map, so resolution fails before any adapter call. Wikilink clicks in files that live entirely inside vault root (e.g. `[[README.md]]`) work because the root is always visible.

This is consistent with the existing note in the docs about plugins that bypass the adapter (Omnisearch) seeing an empty local directory — but in this case the missing piece is Obsidian's own link resolution, not a third-party indexer.

## Workaround (current)

- Pre-expand all subfolders in the File Explorer once per session (tedious, defeats the purpose of a hub page).
[...]

PR #549, opened 2026-09-23
release: 1.1.8 (stable)

sotashimozono:
Stable promotion staging bump. Version-only — no code change.

`main` only ever merges from `next`, so the stable bump lands on `next` first (the `X.Y.Z`-on-`next` exception in `version-check.yml`).

| file | 1.1.8 |
|---|---|
| `manifest.json` (root) | ✅ |
| `manifest-beta.json` | ✅ |
| `plugin/manifest.json` | ✅ |
| `plugin/package.json` + lock | ✅ |
| `versions.json` / `plugin/versions.json` | `"1.1.8": "1.5.0"` added |

### Links, graph and search see the whole vault — #468

The connect walk registers only the root level, and Obsidian resolves links only against `vault.fileMap`. Everything below depth 1 was absent until the user happened to click the folder open, so a `[[link]]` into an unexpanded subfolder silently failed, the graph was partial, and Quick Switcher could not reach a note. `BackgroundIndexer` now walks the whole tree in the background after connect. **This is the fix for #519** (a root hub note linking into nested sub-repos), pinned by a test in #544 in the reported shape: `[[a/b/c/deeper.md|alias]]`, three folders down, beside a `.git/`, with no folder ever expanded.

### A restart no longer re-reads the whole vault — #547 (#513)

Obsidian runs `metadataCache.initialize()` right after every plugin's `onload` and deletes every cached note that is not in the vault model at that moment — and the shadow disk holds none. So each start threw the index away and read every note again over SSH. The plugin now saves the tree (paths, mtime, size — never content) per device and restores it inside `onload`, then reconciles it against the remote after connect.

Measured on 10,000 notes / 50 MB (`plugin/e2e/scale.spec.ts`): **466 s and 10,001 reads → 14 s and 0 reads**, 75 MB → 0.8 MB.

### Every RPC round trip was paying a 40 ms stall — #548

Both ends wrote a frame as two writes (header, then body). Over SSH each write is its own packet, and the TCP under a non-tty session has Nagle on, so the body waited for the peer's delayed ACK. One write per frame: **41 ms → under 1 ms** per round trip, measured from Node against the same daemon.

### Show selected dot-folders — #535, by @JacobStephens2 (#534)

Per-profile **Allowed hidden directories**: exact vault-relative dot-folder paths that should appear in the vault (`.herdr`, `projects/.notes`). Dot-names stay hidden by default, `.obsidian` and ignored dirs can never be re-exposed, and the SFTP walk now prunes excluded dirs before descending.

### Also

- **#467** — stale reads on same-second edits (SFTP reports mtime at 1 s resolution, so the adapter now revalidates on size too); uninstall propagates.
- **#541** — the raw-`fs` work for #429 (#481–#488) is reverted. Patching `fs` in the renderer cannot reach a child process, which is what Claudian is. Raw `fs` under `basePath` is a documented limitation; the six defect specs stay as `test.fixme` with their assertions unweakened.
- **#542 / #543** — CI on Go 1.26 (1.26, not 1.27: 1.27 drops macOS 12, and the daemon runs on the remote) and Node 24 LTS with vitest 5 / jsdom 30. Fixes the `SBOM — server` job that was red on every PR.
- **#463 / #464 / #470 / #471 / #545 / #546** — the E2E suite: sharded, a hard matrix over a deep vault, and a scale benchmark from 1k to 50k notes over a shaped link. #546 also fixed a harness bug that had every launch mint a new vault id — which is what made #513 look like "Obsidian refuses to reuse its index".
- Dependency updates across the plugin and the server.

## Test plan

- [ ] CI green on this PR (commitlint, version-check's promotion-staging shape, lint, type-check, tests)
- [ ] After merge: `gh pr create --base main --head next` to promote, then `release.yml` publishes v1.1.8 from the `main` push


sotashimozono:
Superseded: this was cut when `next` was at beta.21, and the release review that followed changed what 1.1.8 contains (#550–#555). Re-cut from the current `next`.

